PR #37, opened 2017-11-02
Add MPL header to files

joeyg:
Fixes issues #26. I also updated the .gitignore file, let me know if that should be moved to a different PR. Both where taken from the firefox-ios project.


sashei:
Thanks, looks good!

PR #52, opened 2017-11-17
FxA OAuth flow

### LINKED ISSUE #31: OAuth flow with scoped key

sashei:
resolves #31

### DIFF
--- a/lockbox-ios/FxA/FxAPresenter.swift
+++ b/lockbox-ios/FxA/FxAPresenter.swift
@@ -28,7 +28,7 @@ class FxAPresenter {
 
     private var jwkKey:String {
         get {
-            return keyManager.generateRandomECDH().base64URL()
+            return keyManager.getEphemeralPublicECDH().base64URL()
         }
     }
     private var codeChallenge:String { get { return codeVerifier.sha256withBase64URL()! } }
--- a/lockbox-ios/FxA/FxAPresenter.swift
+++ b/lockbox-ios/FxA/FxAPresenter.swift
@@ -21,6 +21,7 @@ class FxAPresenter {
     private var scopedKeySubject = PublishSubject<OAuthInfo>()
 
     internal let oauthHost = "oauth-scoped-keys-oct10.dev.lcip.org"
+    internal let profileHost = "latest-keys.dev.lcip.org"
     internal let redirectURI = "lockbox://redirect.ios"
     internal let clientID = "98adfa37698f255b"
     internal var state:String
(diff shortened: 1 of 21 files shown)


sashei:
not ready for review or merge yet

lockbox-ios/FxA/FxAPresenter.swift
```diff
@@ -0,0 +1,178 @@
...
+}
+
+class FxAPresenter {
+    weak var view:FxAViewProtocol?
+    private var keyManager:KeyManager
+    private var session:URLSession
+    private var scopedKeySubject = PublishSubject<OAuthInfo>()
+
+    internal let oauthHost = "oauth-scoped-keys-oct10.dev.lcip.org"
+    internal let redirectURI = "lockbox://redirect.ios"
```
rfk: Sorry for the drive-by, but I noticed this redirectURI wile reviewing the FxA-side code and I wanted to flag it for follow-up.

One of the issues that came up in the external security review of the scoped-keys work, was that there's a measurable difference in security between using URIs with a custom protocol (like lockbox://redirect.ios) and using https URIs delegated to the native app via universal links (like e.g. https://lockbox.firefox.com/oauth/ios coupled with [1]).  We should aim to use https URIs for redirects wherever possible, because it makes it much harder for a malicious iOS app to try to intercept redirects intended for the lockbox app.

Something to consider for a post-beta follow-up perhaps; I just wanted to put it on your radar while I thought of it :-)

[1] https://developer.apple.com/library/content/documentation/General/Conceptual/AppSearch/UniversalLinks.html

sashei: Nice catch, Ryan. I'm not too worried about the redirectURI for alpha because in the sandboxed WKWebView that I'm using in this implementation, there's no system-level call for any web redirects. We're also using PKCE specifically to protect against the type of malicious redirect interception you mention.

Changing the redirectURI we're using in the sandboxed webview is trivial, though, and would just require a config change in the scoped-keys development environment (@vladikoff might be able to help us out there). If we want to guard against (or support ;) a system-level universal link redirect for something like https://lockbox.firefox.com, there's more work involved but you're right that it's worth looking into for beta.

vladikoff: Yeah just let me know if redirect uri needs to be adjusted

lockbox-ios/Models/KeychainManager.swift
```diff
@@ -0,0 +1,46 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import Security
+
+enum KeychainManagerKey: String {
+    case email, scopedKey
+}
+
+class KeychainManager {
```
linuxwolf: it is important to also save the FxA uid; it is used by `lockbox-datastore` as the salt for deriving the encryption key and hashing key

sashei:
addressed comments; currently, the datastore initialization & hookup will be dependent on other architecture decisions so using the uid & scoped key is not implemented yet (tracked in issue #87)

PR #67, opened 2017-12-06
JS Callback with integration specs

### LINKED ISSUE #65: JavaScript wrapper with callbacks to DataStore code
Right now, it's impossible to chain observables querying .js functions that are async because the evaluateJavascript function does not do async code. 

There needs to be an extension / wrapper for the Datastore class that calls back to the appropriate Swift method to complete observables in the datastore when an async function completes.

Additionally, this will require the datastore initializer to get more robust, as we need to pass each callback to the .js runtime when initializing the webview.

sashei:
resolves #65


sashei:
IRL r+ from @linuxwolf, merging.

PR #84, opened 2017-12-16
Don't require global browserify

joeyg:
Fix the `update-datastore` script to not install browserify globally. On many systems this would cause the script to fail due to write access.


sashei:
r+ LGTM, thanks @joeyg

PR #127, opened 2018-02-22
Use bytesNoCopy:length:encoding:freeWhenDone:

### LINKED ISSUE #120: `Garbage at end` from FxA decryption results in app crash
Sometimes part of the FxA login results in the following error: `Error Domain=NSCocoaErrorDomain Code=3840 "Garbage at end." UserInfo={NSDebugDescription=Garbage at end.}`

As a result, the user is not logged in, and cancelling & trying to login again crashes the app. This happens intermittently, and subsequent login attempts are successful.

#### devices
- iPhone 8 11.2 simulator
- iPhone 5 10.3 simulator

sashei:
import non-null-terminated cstrings safely

resolves #120

### DIFF
--- a/lockbox-ios/Common/Helpers/KeyManager.swift
+++ b/lockbox-ios/Common/Helpers/KeyManager.swift
@@ -18,15 +18,17 @@ class KeyManager {
         return String(cString: jsonValue!)
     }
 
-    func decryptJWE(_ jwe: String) -> String {
+    func decryptJWE(_ jwe: String) -> String? {
         let count = jwe.data(using: .utf8)!.count as size_t
         let contentLen = UnsafeMutablePointer<size_t>.allocate(capacity: count)
         let err = UnsafeMutablePointer<cjose_err>.allocate(capacity: count)
 
         let cJoseJWE = cjose_jwe_import(jwe, count, nil)
-        let decryptedPayload = cjose_jwe_decrypt(cJoseJWE, self.jwk, contentLen, err)
+        guard let decryptedPayload = cjose_jwe_decrypt(cJoseJWE, self.jwk, contentLen, err) else {
+            return nil
+        }
 
-        return String(cString: decryptedPayload!)
+        return String(bytesNoCopy: decryptedPayload, length: contentLen.pointee, encoding: .utf8, freeWhenDone: true)
     }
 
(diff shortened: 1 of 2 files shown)


lockbox-ios/Common/Helpers/KeyManager.swift
```diff
@@ -18,15 +18,17 @@ class KeyManager {
...
-    func decryptJWE(_ jwe: String) -> String {
+    func decryptJWE(_ jwe: String) -> String? {
         let count = jwe.data(using: .utf8)!.count as size_t
         let contentLen = UnsafeMutablePointer<size_t>.allocate(capacity: count)
         let err = UnsafeMutablePointer<cjose_err>.allocate(capacity: count)
 
         let cJoseJWE = cjose_jwe_import(jwe, count, nil)
-        let decryptedPayload = cjose_jwe_decrypt(cJoseJWE, self.jwk, contentLen, err)
+        guard let decryptedPayload = cjose_jwe_decrypt(cJoseJWE, self.jwk, contentLen, err) else {
+            return nil
```
linuxwolf: The only reason `cjose_jwe_decrypt()` returns `NULL` is if it fails.  This might want to "throw" `err` in some form, or at least log it somehow.

## background ##
[`cjose_err`](http://cisco.github.io/cjose/structcjose__err.html) is a structure with a bunch of static values in it, but the most relevant for logs (if nothing else) are:

* `code` describing the error in a machine-readable way
* `message` canned text describing the error for a human (more or less)

lockbox-ios/Common/Helpers/KeyManager.swift
```diff
@@ -18,15 +18,17 @@ class KeyManager {
...
+        }
 
-        return String(cString: decryptedPayload!)
+        return String(bytesNoCopy: decryptedPayload, length: contentLen.pointee, encoding: .utf8, freeWhenDone: true)
```
linuxwolf: CJOSE allows API users to configure what implementation of `malloc` / `realloc` /  `free` is used (by default, whatever it picks up from `libc` or its moral equivalent).

Not knowing enough about Swift, I don't know if there's a concern about deallocation here, but I'm going to assume not if everything passes BuddyBuild (-:

PR #132, opened 2018-03-06
Copy username and password from detail view with confirmation

### LINKED ISSUE #70: Copy / Paste credentials from details view
When I need to access an account, I want to copy the username or password so that I can easily paste those credentials into the native app or mobile browser where I'm logging in.*

(*) Without seamless app or browser integration, copy/paste is the best option for getting a user log in access on iOS.

## Acceptance criteria
- Should be able to copy username from details view
- Should be able to copy password from details view
- Should be informed of copied credential (see https://github.com/mozilla-lockbox/lockbox-extension/issues/318)
- Should be able to navigate out of the Lockbox app, to either app or mobile browser, to paste copied credentials

## Visual design
https://mozilla.invisionapp.com/share/7JEUG47ZU#/screens/267560603

## ToDo
- [x] username field
- [x] password field
- [x] confirmation toast

sashei:
resolves #70


devinreams:
@sashei @changecourse I presume we should inherit the Photon motion curve here? Or is that not as applicable for mobile interaction? https://design.firefox.com/photon/motion/duration-and-easing.html

changecourse:
EASY EASE!

devinreams:
ease

lockbox-ios/Action/CopyAction.swift
```diff
@@ -0,0 +1,46 @@
...
+}
+
+extension CopyAction: Equatable {
+    static func ==(lhs: CopyAction, rhs: CopyAction) -> Bool {
+        return lhs.text == rhs.text &&
+                lhs.fieldName == rhs.fieldName
+    }
+}
+
+struct CopyDisplayAction: Action {
```
linuxwolf: It took me a moment to realize this is the action for showing "Copied to clipboard ✓".  It might be worth future us to have a comment about that here.

linuxwolf: Alternatively, consider a name change to something like CopyConfirmedAction.

PR #145, opened 2018-03-09
filter list of entries

### LINKED ISSUE #46: Search / filter list of saved entries
## Acceptance criteria
- Typing in the search bar dynamically changes the list of entries
- Search should filter the URL, username, and entry/title
- The results should appear in the order that they're returned from the database (no additional support for organizing results at this time)
- the cancel button appears in the search bar when text has been entered
- tapping the cancel button clears the current search text & restores the full list of entries

## Visual Design
**Invision**
Default: https://mozilla.invisionapp.com/share/XCEV238ZD#/267560598_Entry_List
Filtered: https://mozilla.invisionapp.com/share/XCEV238ZD#/283633648_Entry_List_-_Searching

**Zeplin**
Default: https://zpl.io/aBOOd4L
Filtered: https://zpl.io/aRNpKpN

sashei:
resolves #46


lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -38,17 +54,33 @@ class ItemListPresenter {
...
+                    return !latest.items.isEmpty
                 }
+                .map { (latest: ItemsText) -> [Item] in
+                    if latest.text.isEmpty {
+                        return latest.items
+                    }
+
+                    let latestText = latest.text.lowercased()
+                    return latest.items.filter { item -> Bool in
+                        return item.entry.username?.lowercased().contains(latestText) ?? false
```
jhugman: Be really wary about this — if your text is user-enterable, then you can't assume converting to lowercase and comparing will be correct.

Instead, use:
```swift
a.caseInsensitiveCompare(b) == .orderedSame
```

jhugman: I'd also group the optional unwrapping together: 

```swift
guard let itemText = item.entry.username ?? item.origins.first ?? item.title else {
    return false
}
return itemText.caseInsensitiveCompare(latest.text) == .orderedSame
```

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -13,11 +13,23 @@ protocol ItemListViewProtocol: class {
     func hideEmptyStateMessaging()
 }
 
+struct ItemsText {
```
jhugman: Nit: naming. 

Having that `s` in the middle will get old fast.

Suggest, perhaps: `NamedItemList`, or `TextAndItemList`.

lockbox-ios/Storyboard/Base.lproj/ItemList.storyboard
```diff
@@ -77,7 +114,10 @@
                 </viewController>
                 <placeholder placeholderIdentifier="IBFirstResponder" id="WLg-fp-HvJ" userLabel="First Responder" sceneMemberID="firstResponder"/>
             </objects>
-            <point key="canvasLocation" x="750" y="197"/>
+            <point key="canvasLocation" x="749.60000000000002" y="196.55172413793105"/>
```
jhugman: Really?

lockbox-ios/View/ItemListView.swift
```diff
@@ -76,17 +85,44 @@ extension ItemListView: ItemListViewProtocol {
...
-                    guard let cell = tableView.dequeueReusableCell(withIdentifier: "itemlistcell") as? ItemListCell else { // swiftlint:disable:this line_length
-                        fatalError("couldn't find the right cell!")
+        self.dataSource = RxTableViewSectionedAnimatedDataSource<ItemSectionModel>(
+                configureCell: { dataSource, tableView, path, _ in
+                    let cellConfiguration = dataSource[path]
+
+                    var retCell: UITableViewCell
+                    switch cellConfiguration {
+                    case .Search:
+                        guard let cell = tableView.dequeueReusableCell(withIdentifier: "filtercell") as? FilterCell,
```
jhugman: Are these cells `register`ed with the tableView anywhere?

lockbox-ios/View/ItemListView.swift
```diff
@@ -76,17 +85,44 @@ extension ItemListView: ItemListViewProtocol {
...
+                        cell.filterTextField.rx.text
+                                .orEmpty
+                                .asObservable()
+                                .bind(to: presenter.filterTextObserver)
+                                .disposed(by: cell.disposeBag)
+
+                        retCell = cell
+                    case .Item(let title, let username, _):
+                        guard let cell = tableView.dequeueReusableCell(withIdentifier: "itemlistcell") as? ItemListCell else { // swiftlint:disable:this line_length
+                            fatalError("couldn't find the right cell!")
```
jhugman: Ug. Fatal Errors make me queasy.

lockbox-ios/View/ItemListView.swift
sashei: storyboard.

lockbox-ios/Presenter/ItemListPresenter.swift
sashei: `.caseInsensitiveCompare()` does a complete comparison; we should return matches even if the entered text only matches a substring of the username or title or origin.

jhugman: Gah! I misread. 

Nevertheless: we shouldn't be rolling our own case insensitive string ops. They aren't locale aware; there are alphabets that don't have a 1-1 mapping between upper and lower cases. 

Strings are weird. Rely on the OS (via [ICU][1]) to do the work for you.

```swift
guard let itemText = item.entry.username ?? item.origins.first ?? item.title else {
    return false
}
return itemText.localizedCaseInsensitiveContains(latest.text)
```

[1]: http://site.icu-project.org/

PR #154, opened 2018-03-14
Add metrics plan; Define event methods and objects

### LINKED ISSUE #130: iOS Telemetry Metrics Plan
We need to collect event telemetry to tell us the how often and in what ways users are interacting with the iOS app. To this end I (Leif) will prepare a data collection plan that will include the following:

1. Which events will be collected, and when, according the latest invision prototype.
2. Each event's taxonomy: their `category`, `method`, `object`, `value` and `extra` fields. 
3. What other non-event-specific pieces of data need to be attached to an event in its `extra` field, such as a user's fxa uid or random item ids.
4. Why we are collecting these events, and what analyses they will be used for.

The data collection plan will assume usage of [this](https://github.com/mozilla-mobile/telemetry-ios/) library. Where possible, I will give example syntax for event registration and recording.

irrationalagent:
Resolves #130 

OK, here is a first pass at a list of events we'd like to record given the current functionality (as I understand it anyway). In metrics.md I add a few notes of things I'm unsure about; here's a brief tl;dr:

1. We need to have the fxa uid present in the `extras` field of every event where possible (i.e. after authentication) - we currently do this for the extension but if its more difficult here, let me know.

2. I'm unsure of how hard it is to capture errors in a succinct way if fxa authentication fails for some reason. It would be nice to have a short, interpretable error code or message that is added to the extras field on the events that tracks authentication failure / success. 

3. I went ahead and defined the required event methods and objects in `TelemetryIntegration.swift` . If I fouled this up let me know.

Let me know how else I can help or if something is not clear or stupid. I think I get the basics of the mechanics here, so willing to try my hand at placing some of the events in myself, if needed.

Thanks!

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,0 +1,133 @@
+# Lockbox for iOS Metrics Plan
+
+_Last Updated: March 14, 2018_
+
+<!-- TOC depthFrom:2 depthTo:6 withLinks:1 updateOnSave:1 orderedList:0 -->
+
+- [Analysis](#analysis)
+- [Collection](#collection)
+- [List of Proposed Events](#list-of-proposed-events)
+- [References](#references)
+
+<!-- /TOC -->
+
+This is the metrics collection plan for the Lockbox iOS app. It documents all events that are planned to be collected through telemetry. It will be updated periodically to reflect all new and planned data collection.
+
+## Analysis
+
+Data collection is done solely for the purpose of product development, improvement and maintenance.
+
+More here TBD
+
+## Collection
+
+Data will be collected using this library:
+
+https://github.com/mozilla-mobile/telemetry-ios/
+
+We plan to submit two ping types.
+
+First is the "core ping", which contains information about the iOS version, architecture, etc of the device lockbox has been installed on:
+
+https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/core-ping.html
+
+The second is the "focus event ping" which allows us to record event telemetry:
+
+https://github.com/mozilla-mobile/focus-ios/wiki/Event-Tracking-with-Mozilla%27s-Telemetry-Service
+
+The ping types are defined in `lockbox-ios/Common/AppDelegate.swift`. Scheduling of ping transmission is done in the same file.
+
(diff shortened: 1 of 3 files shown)


devinreams:
For waffle/GitHub organization purposes I just edited the original description to include "Resolves #130" ☝️

lockbox-ios/TelemetryIntegration.swift
```diff
@@ -9,14 +9,29 @@ class TelemetryEventCategory {
...
     public static let background = "background"
     public static let foreground = "foreground"
-    public static let click = "click"
+    public static let tap = "tap"
+    public static let signin = "signin"
 }
 
 class TelemetryEventObject {
     public static let app = "app"
     public static let initStoreButton = "init_store_button"
```
sashei: These three lines could be removed as part of this.

lockbox-ios/TelemetryIntegration.swift
```diff
@@ -9,14 +9,29 @@ class TelemetryEventCategory {
 }
 
 class TelemetryEventMethod {
+    public static let startup = "startup"
     public static let background = "background"
```
sashei: I don't see a "backgrounded" or "foregrounded" event in the above enumeration, so these can be removed as well.

docs/metrics.md
```diff
@@ -0,0 +1,133 @@
...
+		* Note: If there is an authentication error, let's put it in the extra field here, if possible.
+
+4. When a user taps an item in the entry list:
+	* `category`: action
+	* `method`: tap
+	* `object`: entryList
+	* `value`: nil
+	* `extras`: ["fxauid" : uid, "itemid" : itemid]
+
+5. When a user taps one of the buttons that are shown in the menu displayed after selecting an entry:
```
sashei: I think that menu is no longer a part of the iOS designs... can @changecourse confirm?

changecourse: Correct... we are removing all functionality around quick-access pane for now, and instead supporting standard chevron, click anywhere on the bar, to access entry details.

irrationalagent:
Thanks for the review, Sasha. I decided to keep the fore/background events after all, as they might be helpful in establishing some general measure of how much time people spend in the app. Not sure what we'd need that for now, but best to have it anyway. I added the spec for it to the md file

sashei:
r+

PR #158, opened 2018-03-14
touch item

### LINKED ISSUE #152: Touch item in DataStore after copy / paste
After a username or password is copied from the app, modify the corresponding `lastUsed:` property on the item in the datastore.

sashei:
resolves #152


lockbox-ios/Action/DataStoreAction.swift
```diff
@@ -273,6 +292,26 @@ extension DataStoreActionHandler {
...
+                    self.checkState()
+                }
+                .flatMap { _ -> Single<Any> in
+                    if item.id == nil {
+                        throw DataStoreError.NoIDPassed
+                    }
+
+                    let jsonItem = try self.parser.jsonStringFromItem(item)
+
+                    return self.webView.evaluateJavaScript("\(self.dataStoreName).touch(\(jsonItem))")
```
jhugman: I… don't… understand… this code. 

Doesn't `evaluateJavascript` return `void`? Oh, god it, Rx monkey patches.

lockbox-ios/Common/Extensions/Date+.swift
```diff
@@ -0,0 +1,16 @@
...
+import Foundation
+
+extension Date {
+    public init?(iso8601DateString: String) {
+        let trimmedIsoString = iso8601DateString.replacingOccurrences(of: "\\.\\d+", with: "", options: .regularExpression)
+        if let date = ISO8601DateFormatter().date(from: trimmedIsoString) {
+            self = date
+        } else {
+            return nil
+        }
```
jhugman: Nit: Could you document some of the sample strings this is expected to handle?

lockbox-ios/Store/DataStore.swift
```diff
@@ -41,6 +45,19 @@ class DataStore {
                     switch action {
                     case .list(let list):
                         self.itemList.onNext(list)
+                    case .updated(let item):
+                        self.itemList.take(1)
+                                .map { items in
+                                    guard let id = item.id else {
+                                        return items
+                                    }
```
jhugman: Ok.

lockbox-ios/lockbox-datastore/dswrapper.js
```diff
@@ -68,4 +68,14 @@ class SwiftInteropDataStore extends DataStoreModule.DataStore {
       }
     })
   }
+
+  async touch(item) {
+    return super.touch(item).then( function(updatedItem) {
```
jhugman: Nit: Use

```js
updatedItem => {}
```

instead of `function`.

lockbox-ios/Common/Extensions/Date+.swift
sashei: Does the test count as documentation here? Or are you saying that I could include more cases in the spec file to be more thorough?

PR #167, opened 2018-03-22
Clear keychain items on first run of app

### LINKED ISSUE #155: Deleting app & reinstalling doesn't require FxA authentication upon opening
Instead of requiring the user to authenticate again when reinstalling the app, it opens the app already on the Entry List screen.

sashei:
resolves #155

### DIFF
--- a/lockbox-ios/Common/AppDelegate.swift
+++ b/lockbox-ios/Common/AppDelegate.swift
@@ -5,6 +5,8 @@
 import UIKit
 import Telemetry
 
+let FirstRunKey = "firstrun"
+
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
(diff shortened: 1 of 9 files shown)


lockbox-iosTests/UserInfoActionSpec.swift
```diff
@@ -0,0 +1,76 @@
...
+                    expect(UserInfoAction.scopedKey(key: "something")).to(equal(UserInfoAction.scopedKey(key: "something")))
+                    expect(UserInfoAction.scopedKey(key: "something")).notTo(equal(UserInfoAction.scopedKey(key: "somethingElse")))
+                }
+
+                it("profileInfo actions are equal when the infos are equal") {
+                    expect(UserInfoAction.profileInfo(info: ProfileInfo.Builder().uid("meh").build())).to(equal(UserInfoAction.profileInfo(info: ProfileInfo.Builder().uid("meh").build())))
+                    expect(UserInfoAction.profileInfo(info: ProfileInfo.Builder().uid("blah").build())).notTo(equal(UserInfoAction.profileInfo(info: ProfileInfo.Builder().uid("meh").build())))
+                }
+
+                it("profileInfo actions are equal when the infos are equal") {
```
linuxwolf: s/profileInfo/oauthInfo/ ?

lockbox-ios/Common/AppDelegate.swift
```diff
@@ -17,6 +19,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         self.window?.rootViewController = RootView()
         self.window?.makeKeyAndVisible()
 
+        if UserDefaults.standard.string(forKey: FirstRunKey) == nil {
```
linuxwolf: I stumbled a little over this.  The flag is **not** set _until_ first run completes. 

I can't think of a better naming convention here, but it might be worth a comment for future us.

PR #168, opened 2018-03-26
Auto lock Setting

### LINKED ISSUE #80: Biometric Access Timer
Provide ability for the user to adjust the length of timer for biometric access. Previously named autolock settings, but that was removed from the intent which is the "ability to unlock with biometrics"

## Acceptance criteria
- Should default to being on
- Should default to 5 minutes until the user sets the timer
- Should be able to set via the settings page (onboarding is different story)
- Should be able to unlock from the same UI as signing in for the first time (until there is a use case for those to be different)

NOTE: This task includes building the UI and autolock setting.

- Dependent on #35
- Dependent on #77

## Visual Design

**InVision**
https://mozilla.invisionapp.com/share/7JEUG47ZU#/screens/267560622

**Zeplin**
https://zpl.io/bAGeO9o

joeyg:
Resolves #80. 

simulator screen shot - iphone x - 2018-03-26 at 12 51 47


lockbox-ios/Presenter/AutoLockSettingsPresenter.swift
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    private func updateAutoLockValue(_ setting: AutoLockSetting?) {
+        let sectionModel = settings.value[0]
+        let newItems = sectionModel.items.map { (cellConfiguration) -> CheckmarkSettingCellConfiguration in
+            cellConfiguration.isChecked = (cellConfiguration.value as? AutoLockSetting) == setting ? true : false
+            return cellConfiguration
+        }
+
+        settings.value = [AnimatableSectionModel(model: 0, items: newItems)]
```
joeyg: @sashei - Not sure if this is the most SwiftRX way. Could the value be modified instead of replaced?

lockbox-ios/View/AutoLockSettingsView.swift
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+extension AutoLockSettingsView {
+    private func setupNavbar() {
+        navigationItem.title = Constant.string.settingsAutoLock
+    }
+
+    private func setupDataSource() {
+        dataSource = RxTableViewSectionedReloadDataSource(
```
sashei: `self.dataSource`

lockbox-ios/View/AutoLockSettingsView.swift
```diff
@@ -0,0 +1,101 @@
...
+}
+
+extension AutoLockSettingsView: AutoLockSettingsProtocol {
+    func bind(items: SharedSequence<DriverSharingStrategy, [AutoLockSettingSectionModel]>) {
+        guard let dataSource = self.dataSource else {
+            fatalError("datasource not set!")
+        }
+
+        items
+            .drive(self.tableView.rx.items(dataSource: dataSource))
```
sashei: `self.dataSource`

lockbox-ios/View/AutoLockSettingsView.swift
```diff
@@ -0,0 +1,101 @@
...
+
+            cell.accessoryType = cellConfiguration.isChecked ?
+                UITableViewCellAccessoryType.checkmark : UITableViewCellAccessoryType.none
+
+            return cell
+        })
+    }
+
+    private func setupDelegate() {
+        tableView.rx.modelSelected(CheckmarkSettingCellConfiguration.self)
```
sashei: `self.tableView`

lockbox-ios/View/AutoLockSettingsView.swift
```diff
@@ -0,0 +1,101 @@
...
+            .subscribe(onNext: { model in
+                guard let val = model.value as? AutoLockSetting,
+                      let presenter = self.presenter else { return }
+
+                if let indexPath = self.tableView.indexPathForSelectedRow {
+                    self.tableView.deselectRow(at: indexPath, animated: true)
+                }
+
+                presenter.itemSelected(val)
```
sashei: This should be bound to an observer on the presenter, not a function call (example in `ItemList` presenters & views for the `itemSelected` observer)

lockbox-ios/Presenter/AutoLockSettingsPresenter.swift
```diff
@@ -0,0 +1,66 @@
...
+import RxDataSources
+
+class AutoLockSettingsPresenter {
+    private var view: AutoLockSettingsProtocol
+    private var userInfoStore: UserInfoStore
+    private var routeActionHandler: RouteActionHandler
+    private var userInfoActionHandler: UserInfoActionHandler
+    private var disposeBag = DisposeBag()
+
+    var settings = Variable([AnimatableSectionModel(model: 0, items: [
```
sashei: Nice use of `Variable`; however, it's being deprecated in future releases of RxSwift, so for future `Variable`s use `ReplaySubject` with a buffer size of 1.

lockbox-ios/Presenter/AutoLockSettingsPresenter.swift
```diff
@@ -0,0 +1,66 @@
...
+        self.routeActionHandler = routeActionHandler
+        self.userInfoActionHandler = userInfoActionHandler
+
+        self.userInfoStore.autoLock.subscribe(onNext: { setting in
+            self.updateAutoLockValue(setting)
+        }).disposed(by: disposeBag)
+    }
+
+    func onViewReady() {
+        let driver  = settings.asDriver()
```
sashei: Rather than binding the `Variable` here, the subscription to `self.userInfoStore.autoLock` should happen in `onViewReady()` & the configuration generated in the observable chain.

lockbox-ios/Presenter/AutoLockSettingsPresenter.swift
sashei: The value will always be replaced when generating a new configuration; that's a pretty Rx-y paradigm (no mutable state etc). Holding on to a `Variable` for this configuration is an unnecessary overhead though; all of the work to generate the configurations should happen inline with the call to `self.userInfoStore.autoLock`.

lockbox-ios/Presenter/SettingsPresenter.swift
```diff
@@ -69,6 +69,11 @@ class SettingsPresenter {
         userInfoActionHandler.invoke(.biometricLogin(enabled: isOn))
     }
 
+    func onItemSelected(setting: SettingCellConfiguration) {
+        guard let routeAction = setting.routeAction else { return }
```
sashei: This should be an `Observer`.

lockbox-ios/Store/UserInfoStore.swift
```diff
@@ -34,6 +35,10 @@ class UserInfoStore {
         return _biometricLoginEnabled.asObservable()
     }
 
+    public var autoLock: Observable<AutoLockSetting?> {
+        return _autoLock.asObserver()
```
sashei: `asObservable()`

lockbox-ios/View/AutoLockSettingsView.swift
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        setupNavbar()
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        setupDataSource()
```
sashei: `self.` for all of these calls.

lockbox-ios/View/SettingsView.swift
```diff
@@ -63,6 +63,7 @@ class SettingsView: UITableViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         setupDataSource()
```
sashei: `self.` for all of these calls.

lockbox-ios/View/SettingsView.swift
```diff
@@ -101,10 +102,24 @@ extension SettingsView {
...
+    private func setupDelegate() {
+        guard let presenter = presenter else { return }
+        tableView.rx.modelSelected(SettingCellConfiguration.self)
+            .subscribe({ evt in
+                if let indexPath = self.tableView.indexPathForSelectedRow {
+                    self.tableView.deselectRow(at: indexPath, animated: true)
+                }
+
+                guard let setting = evt.element else { return }
+                presenter.onItemSelected(setting: setting)
```
sashei: again, bind to the observer on the presenter instead of making a function call.

lockbox-ios/View/SettingsView.swift
```diff
@@ -149,3 +164,14 @@ class SwitchSettingCellConfiguration: SettingCellConfiguration {
         self.isOn = isOn
     }
 }
+
+class CheckmarkSettingCellConfiguration: SettingCellConfiguration {
+    var isChecked: Bool = false
+    var value: Any?
```
sashei: `value` is not a descriptive variable name. I have no idea what that property is for.

lockbox-ios/View/AutoLockSettingsView.swift
joeyg: I think this use of `dataSource` without `self` is because it is defined in the `guard` line above

lockbox-ios/View/AutoLockSettingsView.swift
```diff
@@ -0,0 +1,101 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import UIKit
+import RxSwift
+import RxCocoa
+import RxDataSources
+
+protocol AutoLockSettingsProtocol {
```
sashei: All of the other view protocols a) have `View` in the name and b) are declared in the `Presenter`.

lockbox-ios/View/AutoLockSettingsView.swift
sashei: you're right, my B

lockbox-iosTests/SettingsPresenterSpec.swift
```diff
@@ -31,9 +31,11 @@ class SettingsPresenterSpec: QuickSpec {
 
     class FakeRouteActionHandler: RouteActionHandler {
         var routeActionArgument: RouteAction?
+        var invokeWasCalled = false
 
         override func invoke(_ action: RouteAction) {
             self.routeActionArgument = action
+            self.invokeWasCalled = true
```
sashei: nit: this is extraneous, you can just check that the routeActionArgument isn't nil

lockbox-ios/View/RootView.swift
```diff
@@ -98,6 +98,10 @@ class RootView: UIViewController, RootViewProtocol {
         }
     }
 
+    var isPresentingModal: Bool {
```
sashei: it looks like you wrote this but don't use it anywhere?

joeyg:
@devinreams let me know if there are any other changes needed

devinreams:
Awesome work! Thank you @joeyg  🙇

PR #169, opened 2018-03-28
Load FxA sign in mode not default sign up mode

devinreams:
This assumes the user already has a Firefox Account and takes them straight to sign in (versus tapping out of registration mode).

screen shot 2018-03-28 at 10 04 58 am


sashei:
r+

devinreams:
Thanks for the help and review @sashei.

Selfishly, this will save me ~1 second and 1 tap while end-to-end testing. 😉

PR #184, opened 2018-04-02
UI and spacing improvements to list view

### LINKED ISSUE #157: Entry List adjustments based on designs
Just a couple minor tweaks:

- [x] The bottom entry list item should have a bottom border/separator that extends full width. (note: for technical simplicity, removed bottom which allowed the filter bottom border to extend as expected)
- [x] Filter bottom border should extend full width 
- [x] Looks to be a bit more padding above each Entry Name in list compared to the bottom padding.
- [x] Page background color should be: **#EFEFF4**
- [x] Padding to right of clear button on filter should be **16px**

## Visual Design

**Zeplin**
https://zpl.io/aBOOd4L

Current build screenshot:
list

filter

filter

devinreams:
Resolves #157 

This should address all the items in the issue ☝️ above

screen shot 2018-04-02 at 12 49 39 pm

- [x] I read some articles and played around with the way the separators are inserted. The simplest thing to get the desired effect was make it so the footer item does _not_ have a border (by setting a height), adjust the insets on the cells, and remove the inset on the overall table (so we get a full width border separating the filter). How does everything look to you @nickbrandt?
- [x] I don't really know my way around interface builder but I think I got these addressed as simply/concisely/cleanly as possible. Requesting @sashei review and confirm and make sure any other code stuff I'd need to do is done.

### DIFF
--- a/lockbox-ios/Common/Resources/Constants.swift
+++ b/lockbox-ios/Common/Resources/Constants.swift
@@ -13,12 +13,13 @@ struct Constant {
 
     struct color {
         static let lightGrey = UIColor(hex: 0xEFEFEF)
+        static let listBackground = UIColor(hex: 0xEFEFF4)
         static let lockBoxBlue = UIColor(hex: 0x0088D9)
         static let lockBoxTeal = UIColor(hex: 0x00BAD6)
         static let kebabBlue = UIColor(hex: 0x003EAA)
         static let settingsHeader = UIColor(hex: 0x737373)
         static let settingsBackground = UIColor(hex: 0xEFEFEF)
-        static let tableViewCellHighlighted = UIColor(hex: 0xe5eff9)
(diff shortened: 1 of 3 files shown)


nickbrandt:
@devinreams From what I can see it looks good. Can't seem to get the build to install on my phone but the screenshot looks correct (can't really see borders but from your comments that sounds right). Thanks for knocking these out!

devinreams:
>For instances where you are filtering and there are no results, the border between the filter and list area is lost, and replaced with a white line. Ideally this would remain the gray filter border color 

Good catch @nickbrandt. I updated the footer view color so it's not white:

screen shot 2018-04-03 at 5 18 35 am

screen shot 2018-04-03 at 5 18 39 am


Want to take one final look at this before we call it "done"?

And I see @changecourse opened an issue for an improved (non-empty) empty state: #186

PR #185, opened 2018-04-02
Spacing and size improvements to entry details

### LINKED ISSUE #156: Entry Detail adjustments based on design and copy edits to extension
A few minor tweaks based on how things appear in the designs vs. what has been implemented:

- [ ] Notes field font size appears to be **17px** vs. **14px**
- [ ] Padding between fields/groups as well as between navigation bar and first field is larger than designs (20px)
- [ ] **Web Address** label has been modified to **Website Address** in extension

devinreams:
Resolves #156 

iphone-entry-details

- [ ] I believe I addressed all the items in ☝️ the original issue. I kept the "< Back" button where it is because it is the same alignment as the "A-Z" on the home screen (so the label and images stay in the same place). Requesting @nickbrandt trying that out and also checking the other items.

- [ ] I did some code here, and it seems to work. I think I need tests here? Requesting @sashei review to help explain what else I need to do to wrap this up. 🙃 

Notes:

- I also moved the note about editing in the extension up. I realize that copy needs to change based on newer designs anyway?
- I changed the "A-Z" sorting button and "Back" button to consistently use the system gray and apply it to the image. Otherwise highlight state was a little inconsistent (down arrow was white on highlight and just the text changed grey). I tried our constant lightGrey but it seemed _too_ light and subtle.


devinreams:
@nickbrandt @changecourse I adjusted the highlight state to 60% alpha. If I turn on the very binary "lighten up the image" effect it goes very dark/gray compared and may look better off (like seen here):

highlighted

PR #191, opened 2018-04-03
note the troubleshooting steps if carthage fails on Swift version mis…

linuxwolf:
…match

In case this happens to others less familiar with Swift dev.


devinreams:
Nice 👍 

Small note in case its useful: when I ran into this I _think_ I just re-ran bootstrap and everything worked A-OK. That's kinda nice because it's easy to remember and repeat (ie: we can consistently tell folks to run `bootstrap` kinda like `npm install` on a Node project and things _should_ "just work").

PR #218, opened 2018-04-08
Settings Web Views

### LINKED ISSUE #78: Browser-based info views (settings)
- [x] provide feedback
- [x] FAQ

## Visual Designs
note: really only showing the header bar intended for app content, the page content will be served by the website.

**InVision**
FAQ: https://mozilla.invisionapp.com/share/XCEV238ZD#/267560614_FAQ
Provide Feedback: https://mozilla.invisionapp.com/share/XCEV238ZD#/267560613_Provide_Feedback

**Zeplin**
FAQ: https://zpl.io/25gWjPj
Provide Feedback: https://zpl.io/2yBno3K

joeyg:
Resolves #78.

I used the URLs from the firefox extension but I imagine those will change.

### DIFF
--- a/lockbox-ios/Common/Resources/Constants.swift
+++ b/lockbox-ios/Common/Resources/Constants.swift
@@ -9,7 +9,7 @@ import UIKit
 struct Constant {
     struct app {
         static let redirectURI = "https://mozilla-lockbox.github.io/fxa/ios-redirect.html"
-        static let faqURL = "https://mozilla-lockbox.github.io/lockbox-extension/faqs/"
+        static let faqURL = "https://lockbox.firefox.com/faq/"
         static let provideFeedbackURL = "https://qsurvey.mozilla.com/s3/Lockbox-Input"
     }
 
--- a/lockbox-ios/Common/Resources/Constants.swift
+++ b/lockbox-ios/Common/Resources/Constants.swift
@@ -9,6 +9,8 @@ import UIKit
 struct Constant {
     struct app {
         static let redirectURI = "https://mozilla-lockbox.github.io/fxa/ios-redirect.html"
+        static let faqURI = "https://mozilla-lockbox.github.io/lockbox-extension/faqs/"
+        static let provideFeedbackURI = "https://qsurvey.mozilla.com/s3/Lockbox-Input"
     }
 
     struct color {
--- a/lockbox-ios/Common/Resources/Constants.swift
+++ b/lockbox-ios/Common/Resources/Constants.swift
@@ -9,8 +9,8 @@ import UIKit
 struct Constant {
     struct app {
         static let redirectURI = "https://mozilla-lockbox.github.io/fxa/ios-redirect.html"
-        static let faqURI = "https://mozilla-lockbox.github.io/lockbox-extension/faqs/"
-        static let provideFeedbackURI = "https://qsurvey.mozilla.com/s3/Lockbox-Input"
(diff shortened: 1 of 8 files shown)


lockbox-ios/View/SettingsWebView.swift
```diff
@@ -0,0 +1,38 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import WebKit
+import UIKit
+
+class SettingsWebView: UIViewController {
```
sashei: nit: those extra `s`s sneak up on you.... should just be `SettingWebView`. please change the file name as well when you change this.

lockbox-ios/Common/Resources/Constants.swift
```diff
@@ -9,6 +9,8 @@ import UIKit
 struct Constant {
     struct app {
         static let redirectURI = "https://mozilla-lockbox.github.io/fxa/ios-redirect.html"
+        static let faqURI = "https://mozilla-lockbox.github.io/lockbox-extension/faqs/"
```
sashei: these are called URLs everywhere else you use them :p

devinreams:
>adding @sandysage as a reviewer to update the URLs when we have them.

@sandysage Are you OK with:

1. using https://lockbox.firefox.com/faq/ as the FAQ URL for now?
    - We can always go back and change this when we work on https://github.com/mozilla-lockbox/mozilla-lockbox.github.io/issues/46
2. using the [existing survey URL](https://qsurvey.mozilla.com/s3/Lockbox-Input) for "feedback" for now?
    - Note: It may be worth creating a pre-launch issue to consider gathering iOS specific feedback through a separate survey, perhaps?

nickbrandt:
@devinreams I think that makes sense to keep things moving. The original intent (which probably wan't explained in the issue) was for these external links for FAQ and Provide Feedback to transition in as a "modal" over top the Settings screen (as seen in InVision: https://mozilla.invisionapp.com/share/BQEUSO0S2#/267560609_Settings_-_Touch_ID_Enabled). This is why it makes more sense for this action to be named "Close", but I think it is fine adding a new issue to address this.

sashei:
@nickbrandt yeah, I don't know if that was represented in the Zeplin designs. I also would push back on that interaction paradigm for these views; especially with the disclosure indicator in the settings list I think the stack behavior (pushing on from the right) is to be expected. Adding more modally-presented (up from the bottom on top of other views) views also adds a lot more complexity to the view hierarchy programmatically.

nickbrandt:
@sashei For interactions, I would default to using InVision as the source of truth, as we set up those transitions as intended for development. As for whether or not it should be displayed modally vs. pushing to the left, I can direct you to a few different areas in iOS settings where although they use the disclosure indicator, they still open the screen in a modal (as it is bringing in an external resource). This same behavior also would apply to the FxA sign-in screens, where we would want to instead bring in a modal from the bottom vs. pushing to the left (similar to how Pocket handles signing in with FxA). Happy to discuss this further with @changecourse and @devinreams if it makes sense for us all to come together on this.

sashei:
@nickbrandt sure thing, I will keep an eye on Invision for future transitions. I trust your expertise about the right way to do interactions (it's not like I have an iPhone), so if you want to make a story about fixing those particular pieces of design debt (maybe even including the FxA one) we can figure out the best way to make it work programmatically.

Issue #211, opened 2018-04-05
Address iTunes Connect "issues"


devinreams:
See email from iTunes Connect via Stefan.

- [ ] invalid binary architecture
- [ ] launch storyboard not found
- [ ] invalid bundle
- [ ] invalid Rx segment alignments
- [ ] invalid encryption info
- [ ] non-PIE Binaries

devinreams:
I'm going to remove my assignment here because I think I've exhausted my (in)ability to help here so this is likely something @sashei or @jhugman will have to help with.

Low priority for now compared to feature work, but something I'd like to get proven out (we can build/submit to iTunes Connect) soon so we're not scrambling at the last minute...

PR #222, opened 2018-04-10
Add launch screen storyboard and update welcome screen text

### LINKED ISSUE #212: Add "Launch" screen iOS app assets and storyboard
Aka Loading Screen, Aka Splash Screen

## Visual Design
**InVision**
Splash screen: https://mozilla.invisionapp.com/share/BQEUSO0S2#/267561207_Loading_Screen

**Zeplin**
Splash screen: https://zpl.io/begYxoD

devinreams:
References #211 
Fixes #212 

- [x] add new storyboard to be used for Launch
- [x] add Lockie and ocean (non-gradient) to Launch screen
  - [x] add static logo ~and tagline~ to Launch screen
- [x] add latest logo (Fira font) and tagline to Welcome screen
- [x] fix black loading screen


devinreams:
This is a fun one. Part of the point of the launch screen is it doesn't have all the app (logic) loaded up so stuff like the gradient isn't available yet.

For now I put a _very simple_ nessie + lockbox blue and then the _rest_ of the launch screen loads in: text, buttons, etc. So if you've never signed in before, you see this (pardon the gif quality):

launch-screen

And if the user has logged in before and has data:

launch-screen-signed-in

I'm not sure why we get a black screen inbetween. Do you know @sashei?

Given some of the constraints how does this look @mozilla-lockbox/ux? Do you want to play around a bit more with this together given the blue gradient is not as simple to get in place on launch? And that the screen will show before a first-run (not signed in) _and_ someone who has data. You'd like the "Lockbox" and "tagline" to appear to beneath either way, right?

nickbrandt:
@devinreams Thanks Devin. From what I'm aware design wasn't designated as a reviewer for the PR which contained the initial logo and tagline. I did notice it after the fact, but knew both the logo and tagline was posed to change down the road and were basically placeholders for the time being (still the case). I'll be sure to call things out in the future as design becomes more involved in PRs.

devinreams:
>From what I'm aware design wasn't designated as a reviewer for the PR which contained the initial logo and tagline. I did notice it after the fact, but knew both the logo and tagline was posed to change down the road and were basically placeholders for the time being

OK, that's a good call out. Two ideas:

1. Going forward: I'll make sure we do both code and UX reviews. I'll expect @sashei to help make sure these happen, too.

2. Looking backwards: we likely need you to do a design-to-actual "reconciliation" to show which design/polish items are missed (bugs) so we can prioritize the work either before/during/after the P1 items.

We're certainly open to other ideas but I think that'll keep us moving along. 🚂

devinreams:
>>Ideally, the logo and tagline would be on the launch screen as well. Is this a limitation we have?

>OK, that should also be possible

I shouldn't have said that. The custom font is not available to the app yet. So I believe an image here is  an alternative approach. Or no text at all (how I've left it currently).

I've updated the Welcome screen to use the fonts, sizes, and approximate spacing seen in Zeplin.

updated-fonts

Can you please take a look and let us know what you'd like to see changed?

@sashei said she's looking into the black screen effect.

devinreams:
>List while opening so that app doesn't hang

Thanks @sashei, thats working great for me now.

@mozilla-lockbox/ux I suppose last up is getting a static image loaded onto the "Launch" screen since it doesn't have access to the Fira font like the 'Welcome' screen does. Can you please provide that given I'm using the flat blue (unable to do gradient consistently) on this screen?

Let me know if you want to talk through some options here a bit more..

nickbrandt:
@devinreams You can download the PNGs from Zeplin now: https://zpl.io/begYxoD
Let me know if you need anything else

devinreams:
>You can download the PNGs from Zeplin now
>Let me know if you need anything else

@nickbrandt Thanks. I loaded them in and now we're ready for your official UX/UI review of all the items in the checklist up above ☝️ 

screen shot 2018-04-11 at 11 43 42 am
screen shot 2018-04-11 at 11 43 48 am

devinreams:
>The only question I have is whether or not this is supposed to cover the change in the tagline for the Welcome page to be in italics as well (in latest build it loads with italics and transitions to regular)

Ah, that's a bug.

When I changed over to a different way to display text (Attributed vs Plain) so we could do 1.15 line spacing, the custom font disappeared. This seems to be an issue with Apple Interface Builder: https://openradar.appspot.com/18716979

Fastest solution is to remove the line spacing and update the image so the Launch screen matches the "Plain" treatment. Or I can play with a code solution to get the font loading/replacing with the proper line spacing on the Welcome screen. Want to go with the former first and I'll file an issue to tackle the latter potentially later?

nickbrandt:
@devinreams 👍 That sounds good to me

devinreams:
After playing with this more I still can't quite get the tagline to line up with proper line spacing. Chatted with @nickbrandt and we'll drop the tagline for now and continue discussing at #236 what we want to do since we need to consider use of image (not using text) means a little harder to translate down the road...

devinreams:
@mozilla-lockbox/ux OK, can you pull this down and check it out one more time to make sure we didn't miss anything on the Launch and Welcome screens? 🥇 

launch-welcome

Issue #239, opened 2018-04-12
Update login button title on welcome screen ("Get Started") when not connected to Firefox Account yet


sashei:
~Currently, the button always says "Get Started". It should say "Sign in with your Firefox Account" on the welcome screen when the application is locked.~

Changed default back to "Sign in with your Firefox Account" at #244.

Re-framed this issue to be a nice-to-have to introduce logic to change the label to "Get Started" when no Firefox Account is connected (primarily experienced during a first run, and after unlinking, both likely to be rare occurrences).

Welcome ("Get Started"): https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5a4ea0625387ae000e1a0780

Returning User ("Sign in with your Firefox Account") https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5ac2933f2cbc9d64992d6273

Related to #223

devinreams:
This is my mistake. I introduced this change at #222 and didn't look at the two states of the button.

1. first time, not signed in: "Get Started"
2. has an account, needs to re-authenticate: "Sign in..."

For purposes of getting this back to more useful/usable and closest to the intended UX/UI I'll change the string back.

Then I think we mark this as a P2/nice to have to insert the logic that changes the button string to "Get Started" if there's no account connected yet.

@sashei does that sound right to you?
@mozilla-lockbox/product do you agree with that proposal?

devinreams:
>That sounds like a good plan to me

Cool, thanks @nickbrandt. I changed the default string back to "Sign in with your Firefox Account" at #244.

I also re-framed this issue to be a nice-to-have to introduce logic to change the label to "Get Started" when no Firefox Account is connected (primarily experienced during a first run, and after unlinking, both likely to be rare occurrences).

Descartess:
I will pick this up

devinreams:
Thanks @Descartess, let us know if you have any questions 🙇

devinreams:
Due to changes introduced at #314 we're not going to have a "Get Started" vs "Sign In" state because "Sign In" is going away (relying entirely on biometrics/device PIN).

This will already be completed at #324

PR #244, opened 2018-04-12
Set default button text to "Sign in.." instead of "Get Started"

devinreams:
I accidentally changed this to "Get Started" but realize that's for first run, not signed in. "Sign In" is the best default and "Get Started" for first run/not signed in is arguably a nice-to-have and rarely encountered.

Related to #239

screen shot 2018-04-12 at 2 18 02 pm


devinreams:
"Someone" asked if this fits on an iPhone SE:

screen shot 2018-04-12 at 2 57 41 pm


It appears to! ✅

PR #247, opened 2018-04-14
deselect settings row after tapped

joeyg:
Remove the cell selection after it was tapped.

### DIFF
--- a/lockbox-ios/View/SettingListView.swift
+++ b/lockbox-ios/View/SettingListView.swift
@@ -58,8 +58,9 @@ extension SettingListView {
         self.dataSource = RxTableViewSectionedReloadDataSource(
                 configureCell: { _, _, _, cellConfiguration in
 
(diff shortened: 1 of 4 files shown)


lockbox-ios/View/SettingListView.swift
```diff
@@ -83,6 +83,7 @@ extension SettingListView {
         if let presenter = self.presenter {
             self.tableView.rx.itemSelected
                     .map { path -> SettingRouteAction? in
+                        self.tableView.deselectRow(at: path, animated: true)
```
sashei: Ideally, this would be done in a way that lets us set the background color for the `highlighted` state. You don't need to manually deselect cells; rather, you can set the `selectionStyle` for the tableview to `None`, and add the `tableViewCellHighlighted` color (from `Constants.swift`) for the highlighted state.

joeyg: Ah thanks! I'll fix that up at lunch.

sashei:
Approach looks good; just needs specs!

PR #250, opened 2018-04-16
Expire the copy pasteboard after 60 seconds

### LINKED ISSUE #232: Timer for Copy Action
Establish a standard, yet to be determined timer for copy actions within the app, so passwords don't stay on the clipboard for perpetuity...

devinreams:
Resolves #232 

This was even easier than I thought so went ahead to attempt doing it myself.

Two things:

1. I got stuck making getting a type set in "setItems" but after crawling around Stack Overflow I think this is right... maybe there's a simpler/better/correct-er way?

2. I know some sort of test/spec is needed but it was not obvious to me how I should wait and make sure the pasteboard was emptied. That part may be harder than the change itself...?

---

- [x] need to confirm "60 seconds" is the right length we want
- [x] need to fix / write tests

### DIFF
--- a/lockbox-iosTests/CopyActionSpec.swift
+++ b/lockbox-iosTests/CopyActionSpec.swift
@@ -19,16 +19,14 @@ class CopyActionSpec: QuickSpec {
     }
 
     class FakePasteboard: UIPasteboard {
-        var newString: String?
+        var passedItems: [[String: Any]]?
+        var options: [UIPasteboardOption: Any]?
 
-        override var string: String? {
-            get {
-                return super.string
-            }
-            set {
-                self.newString = newValue
-            }
+        override func setItems(_ items: [[String: Any]], options: [UIPasteboardOption: Any]) {
+            self.passedItems = items
+            self.options = options
         }
+
     }
 
     private var dispatcher: FakeDispatcher!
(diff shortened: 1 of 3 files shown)


sashei:
@devinreams we inject the pasteboard for testing, so you should just be able to test that the correct parameters get set in the `setItems` call (you will need to manually stub out `setItems`) in `CopyActionSpec.swift`. I think making the timeout a constant in `Constants.swift` would be nice also.

devinreams:
>we inject the pasteboard for testing, so you should just be able to test that the correct parameters get set in the setItems call (you will need to manually stub out setItems)

I know what you're saying but after ~30 minutes of messing around.. I don't really know how to accomplish this. 😔

Any chance you can help me out later on when you're not busy? Pair or patch or otherwise?  

>I think making the timeout a constant in Constants.swift would be nice also.

Fair enough! ✅

Did I need to set a type in the constant definition or is the way its wrapped a good/acceptable way to do that?

lockbox-iosTests/CopyActionSpec.swift
```diff
@@ -19,16 +19,14 @@ class CopyActionSpec: QuickSpec {
...
-                return super.string
-            }
-            set {
-                self.newString = newValue
-            }
+        override func setItems(_ items: [[String: Any]], options: [UIPasteboardOption: Any]) {
+            self.passedItems = items
+            self.options = options
         }
+
```
sashei: 👾

devinreams: :boom:

devinreams:
Code and tests in place and reviewed.

Just waiting on @mozilla-lockbox/product for final approval of the here before merging.

devinreams:
This was actually passing at the time of merge:
https://dashboard.buddybuild.com/apps/5a0ddb736e19370001034f85/build/5ad66628c2ced90001ca2862

(the original build timeout sent a CI build status update to GitHub after the rebuild/green)

PR #257, opened 2018-04-17
Set the status bar color in SettingWebView screens

joeyg:
Sets the status bar color in the FAQ and Provide Feedback screens.


devinreams:
Nice catch 👍

PR #260, opened 2018-04-17
fix itunes connect upload errors

### LINKED ISSUE #211: Address iTunes Connect "issues"

sashei:
resolves #211


sashei:
There are no errors when uploading the build anymore; however, we still need to update the information about the app in iTunes Connect & get an admin on ITC to step through the export compliance flow (I realized this may slow down our reviews as we do use non-standard encryption :p)

PR #269, opened 2018-04-18
Do not open browser window when preferred browser setting changes

### LINKED ISSUE #264: Fix changing preferred browser
Right now ExternalLinkActionHandler subscribes to preferred browser setting changes resulting in the browser opening unexpectedly.

Steps to reproduce
1. Open app and tap on a lockbox list item
2. Go back to the app
3. Go to Settings->Preferred Browser
4. Change browser

Expected Result: The setting changes
Actual Result: The new browser opens with the previously clicked on items

joeyg:
Resolves #264. There was some discussion in the Issue about moving more of the code out of the action handler - I am happy to do that change instead if it is preferred.


lockbox-ios/Action/ExternalLinkAction.swift
```diff
@@ -38,6 +38,7 @@ class ExternalLinkActionHandler: ActionHandler {
 
     private func openUrl(string url: String, application: OpenUrlProtocol = UIApplication.shared) {
```
sashei: unrelated, but while you're in here...

why are you injecting the application twice? it should be sufficient to just use `self.application` in this method, as you've already injected it in `init`

devinreams:
*Super Contributor™ @joeyg finds and fixes his own 1 line bug.. and then writes the 12 lines of tests to go along with it.* 😍 🙏

joeyg:
Sorry it was a bug in the first place!

PR #296, opened 2018-04-24
telemetry refactor, added example for copyaction

sashei:

PR #311, opened 2018-04-26
Update metrics.md; add telemetry events

### LINKED ISSUE #58: Add telemetry events for relevant actions in app
Dependent on #130

..and now #180

### LINKED ISSUE #180: Update metrics.md telemetry based on latest requirements
Once @sandysage has confirmed the requirements are ready, go review and update metrics.md.

- [X] Are you using Google Analytics or Ping Centre? - Telemetry
- [X] Metrics.md in the repo in a docs folder explaining the data being sent by the experiment
- [x] Metrics.md is updated to reflect the mvp requirements once designs are complete
- [ ] Review of metrics.md by a data analyst
- [ ] Data Collection Review from Peer/Adjacent https://wiki.mozilla.org/Firefox/Data_Collection
- ~Define what to do when the DNT flag is set~ (does not apply to mobile)
  - Add ability to disable Telemetry usage reporting at #258

Related to #258
Related to #58
Related to #296

irrationalagent:
Resolves #58 
Resolves #180 

Here's an updated metrics md for Sandy (and Devin). I'll wait till https://github.com/mozilla-lockbox/lockbox-ios/pull/296 is merged, then rebase and add the events in this PR


irrationalagent:
@sandysage if you have comments on the doc let me know and I'll make the changes

irrationalagent:
@sashei was there any particular way that you were able to verify that the (e.g.) `CopyAction` telemetry events were being recorded directly? I tried grepping for it in the simulator data after performing the action but that didn't seem to work

sashei:
@irrationalagent hrm ok, I will investigate.

sashei:
@irrationalagent it looks like we don't collect events for simulator builds; on line 56 in `TelemetryAction.swift`, just switch `isCollectionEnabled` from false to true.

irrationalagent:
looked at that line several times and was like 'nah, that's not it' 🤦‍♂️

sashei:
@irrationalagent ok, I got to fool around with this a little bit before james got here this morning, so I may be wrong about it, but I have a very strong suspicion ;) let me know how you go & I can take a closer look if that's not it.

irrationalagent:
Note: in order to debug these events I changed these lines in TelemetryAction.swift, but reverted the changes before committing here:

```swift
#if DEBUG
        telemetryConfig.isCollectionEnabled = true // this one
        telemetryConfig.isUploadEnabled = false
        telemetryConfig.updateChannel = "debug"
```
```swift
    lazy var telemetryActionListener: AnyObserver<TelemetryAction> = {
            target.telemetry.recordEvent(
                    value: action.value,
                    extras: action.extras
            )
            print (    // I added this print statement
              TelemetryEventCategory.action.rawValue,
              action.eventMethod.rawValue,
              action.eventObject.rawValue,
              action.value,
              action.extras
            );
        }.asObserver()
    }()
}
```
Note that 
(1) I'm not sure if the first change was actually required and 
(2) even though adding the print statement logged the correct event data to xcode console and at the right times, I'm not sure it actually means that the events are being recorded. It could just mean they're getting to the spot at which they are supposed to be recorded (at least IIUC).

irrationalagent:
@sashei no rush on this review. @devinreams if want to take it that's fine too

devinreams:
>no rush on this review. @devinreams if want to take it that's fine too

Thanks, I'll test it out locally and make sure I can reproduce your console logging, at the least.

I'll defer to the experts on the code review itself. In the meantime, a fun photo that I'll think of every time I see this PR number:

irrationalagent:
a couple of other notes:
1. The `revealPassword` event always fires when someone views an item detail, but the `value` field logs whether it was revealed (couldnt figure out how to log it in the reveal-only case)

2. I'm using the `value` field now to log settings values etc, let me know if we need to some kind of validation there

3. I had to change the test that references `faqButton` since that event object doesn't exist anymore

lockbox-ios/Action/FxAAction.swift
```diff
@@ -38,6 +63,39 @@ enum FxAError: Error {
     case Unknown
 }
 
+extension FxAError: TelemetryAction {
```
sashei: The FxA error never gets sent as an `Action`, so this extension won't ever get invoked. I think we may need to circle back with @jhugman about how to do this logging in our current "old"-FxA world.

irrationalagent:
went ahead and removed the fxa bits for now. looks like we *may* need some more tests here too ?

sashei:
@irrationalagent yesss, I'd say so. Every `[something]Action.swift` has a corresponding `[something]ActionSpec.swift` file that you can extend with the `TelemetryAction` bits.

devinreams:
From today's standup, @sashei and @sandysage are re-requested for final review.

Nice work @irrationalagent 👏

lockbox-ios/Action/RouteAction.swift
```diff
@@ -29,6 +29,34 @@ enum MainRouteAction: RouteAction {
...
+        case .detail:
+            return .entryDetail
+        }
+    }
+
+    var value: String? {
+        return nil
+    }
+
+    var extras: [String: Any?]? {
```
sashei: These extras are not covered by any tests.

lockbox-ios/Action/RouteAction.swift
```diff
@@ -38,6 +66,38 @@ enum SettingRouteAction: RouteAction {
...
+            return .settingsPreferredBrowser
+
+        }
+    }
+
+    var value: String? {
+        return nil
+    }
+
+    var extras: [String: Any?]? {
```
sashei: not covered by tests

lockbox-ios/Action/RouteAction.swift
irrationalagent: @sasha how do i do something like: 

```swift
it("returns item id as extra") {
     expect(MainRouteAction.detail(itemId: "sfddfs").extras).to(equal([ExtraKey.itemid.rawValue: "sfddfs"]))
}

**Comment by sashei** [comment]
@irrationalagent maybe try `.to(contain(x))` ?

File: lockbox-ios/Action/RouteAction.swift (hunk 2 above)
Comment (sashei) [reply lockbox-ios/Action/RouteAction.swift]: @irrationalagent maybe try .to(contain(x)) ?

Comment (irrationalagent) [reply lockbox-ios/Action/RouteAction.swift]: 😞 feel defeated by this... I've tried a many things but can't get something to work... especially these seem like they should work but im obviously in the dark when it comes to typing in swift

```swift
expect(MainRouteAction.detail(itemId: "sfddfs").extras).to(equal(["itemid": "sfddfs"]))
```
`Cannot convert value of type 'Predicate<[String : String]>' to expected argument type 'Predicate<[String : Any?]>'`

```swift
expect(MainRouteAction.detail(itemId: "sfddfs").extras!.Values).to(contain("sfddfs"))
```
or
```swift
expect(MainRouteAction.detail(itemId: "sfddfs").extras!.Values).to(equal(["sfddfs"]))
```
`Static member 'Values' cannot be used on instance of type '[String : Any?]'`

I considered converting the schema(?) for extras to be 

```swift 
var extras: [String: String?]? { get }
```

but I'd prefer not to since some extras might be integers in the future and it would involve changing a bunch of other lines as well... plz halp

sashei:
heya @irrationalagent , let me know if this update makes sense to you! swifty silliness :(

irrationalagent:
Ah, thanks @sashei TIL. couldn't quite get there myself, thanks for bringing this PR over the line.

PR #314, opened 2018-04-30
Integrate with sync data

### LINKED ISSUE #283: Change to use ios-fxa-sync for Auth and Sync and Storage
Follow up to #241

jhugman:
This PR is used as a temporary master to coordinate the work of @jhugman and @sasha to integrate 
sync and storage as a Store for logins.

Resolves #283


devinreams:
Great work! I'm so impressed you hooked this up so quickly.

FWIW this is building A-OK for me locally and was able to sign in with a test account (it sent me a confirmation email before completing sign in). ✅ 

---

But, with my personal / real account FxA got stuck at sign in. After entering my personal username/password the form spins and spins and then shows the red "Waiting..." bar on the page. Then never completes (I left it for 20+ minutes). ⚠️ 

I sometimes see this when I use my personal account to sign in using FxA into Pocket (but that's OAuth). So I figured I'd share in case its not just me right now and I've stumbled onto a "less happy path" you're not already familiar with...

---

I can't wait to see what tomorrow brings us! 😀

bootstrap.sh
```diff
@@ -9,5 +9,3 @@ carthage bootstrap --platform iOS --color auto --cache-builds
 cd lockbox-ios
 tar xvfz binaries.tar.gz
 cd ..
-
-source 'scripts/update-datastore'
```
devinreams: Weeeee 🕺

@sashei Should references also be removed from the `buddybuild_prebuild` and `gitignore` files?

sashei:
yesssss!!!! *shoulda grep'ed*

devinreams:
Thank you for the update (in the issue) @sashei and nice work getting back to green ✅

sashei:
✅ yes
☔️ no
😉

devinreams:
I am able to fully lock myself out of my app because biometrics stopped prompting me:

1. from Settings > Lock Now
2. back to Welcome screen, prompted for biometrics: hit Cancel
3. background Lockbox aka "Go Home"
4. open Lockbox: see Welcome screen but no prompt for biometrics ⚠️ 

I'm able to reproduce this consistently on simulator. For step 2 you can also just try to "Go Home" (instead of explicitly "Cancel"ing), the result is the same.

I'm able to recover though by force quitting the app and re-launching.

FWIW, there's currently not a great path for a user to go from Lock Now to home screen because they're immediately prompted for biometrics so they _have_ to cancel before they can move on.

---

Also: the lock timer branch has been merged #326 so this is ready for rebase/merge against `master`.

devinreams:
And in testing with iPhone X on simulator, it's no longer prompting me for my face in any combination of things I've tried. It only requests the device passcode.

I did try resetting the device, erase all content and settings, rebuild and reinstall the app, etc.

I also see this behavior on `master` if I don't "Enable" via the onboarding flow, and _even_ if I toggle the setting manually after (it just flips to true, no prompt for face)..

sashei:
@devinreams @nickbrandt I was thinking about it last night & we may want to have a button or something for re-launching biometric auth when you're still on the screen, as well...

devinreams:
@sashei with iPhone X hardware on build 985:

- removed/reinstalled
- signed in with FxA
- Settings > Lock Now 💥 Lockbox crashes ⚠️ 
- re-open Lockbox 💥 black screen and crash ⚠️ 

Locally with iPhone X on Simulator it works A-OK (prompts for pin but I've never seen it actually prompt for face even though it has on the actual phone and when I've "Enrolled" Face ID). ✅ 

And with iPhone 8 on Simulator it seems to hang on the Launch screen with a `SIGABRT` crash on Thread 1? ⚠️ 

screen shot 2018-05-09 at 8 34 18 am


---

Agreed on having the icon ([Face] or Fingerprint) on the screen still so if you cancel you can re-prompt. Are you thinking we file that as a follow-up enhancement?

sashei:
😬 iPhone X struggs. thanks for checking Devin. 🤦‍♀️ usage description......

Yes, I think the button should be a follow-up issue at this point, apologies that it got ripped out in this branch already.

sashei:
I can't replicate the iPhone 8 simulator issue. maybe clean / clear derived data?

PR #326, opened 2018-05-02
Build the auto lock timer

### LINKED ISSUE #214: Autolock timer to lock the app
When I configure the biometric timer, from either onboarding or from settings, I want the app to lock according to my selection.

## Acceptance critiera
- Timer selection should lock the app after that time period configured
- Default of '5 minutes' should lock the app after 5 minutes
- If 'Never' is selected, this configuration should not lock the app
- Timer should lock after each user interaction, which means the app may lock while still open

joeyg:
Resolves #214.


lockbox-ios/Store/AutoLockStore.swift
```diff
@@ -0,0 +1,132 @@
...
+            }
+        default:
+            return
+        }
+    }
+
+    private func setTimer(seconds: Int) {
+        let timerValue = self.userDefaults.double(forKey: SettingKey.autoLockTimerDate.rawValue)
+        if timerValue != 0 {
+            self.timer = Timer(fireAt: Date(timeIntervalSince1970: timerValue),
```
sashei: I think there's an `RxCocoa` binding for `Timer`s so that you can avoid the `@objc` annotation.... might even save you from having to save the reference but I didn't investigate that much.

sashei:
The approach looks good to me!

lockbox-ios/Store/AutoLockStore.swift
joeyg: Do you happen to have a link to that documentation? I am having trouble finding it.

lockbox-ios/Store/AutoLockStore.swift
```diff
@@ -0,0 +1,114 @@
...
+        if let timer = self.timer {
+            timer.invalidate()
+        }
+
+        self.userDefaults.removeObject(forKey: SettingKey.autoLockTimerDate.rawValue)
+    }
+
+    @objc private func lockApp() {
+        self.settingActionHandler.invoke(SettingAction.visualLock(locked: true))
+        self.routeActionHandler.invoke(LoginRouteAction.welcome)
```
sashei: I think the `routeActionHandler` call is extraneous, we do that routing in the `RootPresenter` based on the `visualLock` setting anyway.

lockbox-ios/Store/AutoLockStore.swift
```diff
@@ -0,0 +1,114 @@
...
+            if locked {
+                self.stopTimer()
+            } else {
+                self.setupTimer()
+            }
+        }).disposed(by: self.disposeBag)
+
+        self.dispatcher.register
+            .filterByType(class: LifecycleAction.self)
+            .subscribe(self.lifecycleAction)
```
sashei: whoa, didn't know you could do that! very cool!

joeyg:
@devinreams Thanks for the help testing the feature. Let me know if anything doesn't work as expected and I'll take a look. I mostly focused on restarting the app and making sure it locked/unlocked based on the timer.

devinreams:
>I mostly focused on restarting the app and making sure it locked/unlocked based on the timer.

@joeyg Things are looking pretty consistent and good here. One consideration that may require some changes or discussion:

- set a timer (1 minute)
- "go home" ("close" the app)
- lock my screen
- unlock device after the timer (2+ minutes)

Then the app launches locked as expected. ✅ 

But if I:

- set a timer (1 minute)
- lock my screen (don't home/quit the app)
- unlock device after the time (2+ minutes)

The app opens unlocked instead of locked ⚠️ 

...which is contrary to the product expectation as I read it:

>Timer should lock after each user interaction, which means the app may lock while still open

Because I interpret that description to be _inclusive_ of app backgrounding. Aka after 1 minute, it could lock when user is staring at it and hasn't touched it therefore it could lock if in the foreground but device is locked.

What do you think?

joeyg:
@devinreams ah let me look into getting that fixed.

devinreams:
>The app opens unlocked instead of locked ⚠️

Hmm, I may be crazy but I do _not_ get this behavior if I do "lock on app exit". If I sign in/unlock and immediately lock my device, then unlock my device, Lockbox appears locked. This maybe only happening with timed locks?

joeyg:
@devinreams Sorry about that last diff - I hadn't worked with `Timer` before and I wasn't using it correctly. This one should be working a bit better.

devinreams:
> I hadn't worked with Timer before and I wasn't using it correctly.

Aha! Thanks @joeyg, no apologies needed. This is working great for me!

PR #322, opened 2018-05-01
Rename to Firefox Lockbox

devinreams:
Fixes #323 
Two related items already handled at #320 and #327 

Quick pass at finding and replacing standalone instances of "Lockbox" with "Firefox Lockbox". ~Waiting for final inventory of changes from @mozilla-lockbox/ux.~

Note that the name gets a little awkward when trying to distinguish between the overall product vs this iOS app. In docs I called it "Lockbox Firefox for iOS" for now...

PR #334, opened 2018-05-07
fix specs firefox_lockbox

sashei:
Follow up to #322

### DIFF
--- a/Lockbox.xcodeproj/project.pbxproj
+++ b/Lockbox.xcodeproj/project.pbxproj
@@ -1460,6 +1460,7 @@
 					"${PROJECT_DIR}/lockbox-ios/binaries",
 				);
 				PRODUCT_BUNDLE_IDENTIFIER = "org.mozilla.ios.Lockbox${BUNDLE_ID_SUFFIX}";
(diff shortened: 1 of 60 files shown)


devinreams:
Ugh, I'm super curious as to why these tests didn't fail before landing on master...

sashei:
me too! I went & checked out the BB history & everything, very strange.

sashei:
actually, I think the branches run against the branch target, which you left with the module name Lockbox..... that would make sense.

sashei:
...wait, this means that all the _branches_ will fail with my change here. I think we need them to be unified if we want specs to build on both branches & release builds.

devinreams:
>you left with the module name Lockbox
>we need them to be unified if we want specs to build on both branches & release builds

Oh neat. I see...

I can't quite tell where to make changes to unify. Does it have to do with the product names?

https://github.com/mozilla-lockbox/lockbox-ios/blob/f1c6662fe551284185a076ce5bae79d481aece49/Lockbox.xcodeproj/project.pbxproj#L1571-L1606

devinreams:
>fix module declaration instead

OIC. Sorry, I knew I was going to unexpectedly cause a problem here. 😔

PR #335, opened 2018-05-07
Set FxA redirect_uri back to expected

devinreams:
Follow up to #322 

I'm now getting a "Bad Request: Incorrect redirect_uri" error on the FxA sign in flow. I _must_ have tested signing out and back in when working on that branch but ... perhaps not. 😔

### DIFF
--- a/lockbox-ios/Common/Resources/Constants.swift
+++ b/lockbox-ios/Common/Resources/Constants.swift
@@ -8,7 +8,7 @@ import UIKit
 
 struct Constant {
     struct app {
-        static let redirectURI = "https://lockbox.firefox.com/fxa/ios-redirect.html"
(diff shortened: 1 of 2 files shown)


devinreams:
Tested Sign Out > Sign In > Sign Out > Sign In with Simulator to see if there was some second time sign in problem I missed. Also tried a full hardware reset "Erase all content and settings" (so no existing FxA ID or anything) just to make super sure nothing was left behind and I could try to encounter this again...

I didn't, signed in with no errors. ✅ 

FxA authentication works locally (again).

devinreams:
Also confirmed with Build 941 on my iPhone X hardware, just in case I missed something yet still. I think this bug is properly squashed. Sorry for the interruption.

PR #345, opened 2018-05-10
provide placeholder string when profileinfo not populated

### LINKED ISSUE #342: App crashes on Welcome screen
@sashei I think there's a problem causing crashes due to biometrics.

- iPhone 7: Sign in with FxA, Settings, Lock Now > crash on Welcome (launch app = crash)
- iPhone X: launch app = immediate crash

One theory: since there is no prompt to enable Face/Touch ID at first run its crashing?

Either way, using two test devices, the app is stuck in a crashing / unrecoverable state.

sashei:
resolves #342

### DIFF
--- a/lockbox-ios/Common/AppDelegate.swift
+++ b/lockbox-ios/Common/AppDelegate.swift
@@ -30,6 +30,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         if UserDefaults.standard.string(forKey: PostFirstRunKey) == nil {
             SettingActionHandler.shared.invoke(.reset)
             UserInfoActionHandler.shared.invoke(.clear)
+            DataStoreActionHandler.shared.invoke(.reset)
             UserDefaults.standard.set(false, forKey: PostFirstRunKey)
         }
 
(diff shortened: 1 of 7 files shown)


sashei:
Getting this in to get a fix out for locked-out devices...

@nickbrandt @changecourse open question as to what the placeholder string should be when there is an issue fetching a user's ProfileInfo email address (hopefully this case won't exist when we switch back to OAuth, but for now....)

devinreams:
Aha! Now I’m not crashing on start and it’s clear I don’t have a FxA/Sync account properly hooked up: the list view is empty and Account settings shows the placeholder username@domain.com address. 

If I try to “reset” myself by hitting “Clear out Firefox Lockbox” that does consistently crash, though. 

So even if I delete and re-install I’m still, as far as I can tell, stuck. But so close!

Thanks for the quick work @sashei.

changecourse:
Let's change the placeholder "username@domain.com" to "Firefox Account" if we're unable to pull account information from the library.

devinreams:
Just re-tested on iPhone X hardware (Build 1014) and was able to get out of the broken state:

- reinstalled app
- opened app (I think it crashed but got distracted)
- opened app again or switched to it
- clicked "Get Started" and signed in with FxA 🆗 
- Settings > Accounts had placeholder profile photo and email
- "Clear out Firefox Lockbox" did reset and go back home to sign in
- clicked "Get Started" and signed in with FxA again 🆗 
- manually "Lock Now"ed and was prompted for Face ID permissions 🆗 
- Face ID unlocked and back into entries list ✅ 

That said, when doing smoke tests of the rest of the app:

- Lockbox did automatically lock "on exit" when force quit into background 🆗 
- Lockbox did not automatically lock "on exit" after going home, changing between apps and re-opening (foregrounding related regression?) ⚠️ 
- Lockbox did not automatically lock after 1 minute (timer/related regression?)⚠

PR #346, opened 2018-05-10
loading indication for syncing

### LINKED ISSUE #233: Loading Indication for Initial Sync Import
When loading up my entries for the very first time, I should be given an affordance of something happening within the app, e.g., a loading indication, if it takes more than two seconds to load and import the data into the app.

## Visual Design
**InVision**
https://mozilla.invisionapp.com/share/BQEUSO0S2#/290701009_Entry_List_-_Syncing

**Zeplin**
https://zpl.io/2Ejdl78

sashei:
resolves #233


devinreams:
>loading and enabling working as enumerated

Nice! 👏

lockbox-ios/Common/Extensions/UIViewController+.swift
```diff
@@ -22,35 +23,20 @@ protocol AlertControllerView {
                                 style: UIAlertControllerStyle)
 }
 
+protocol SpinnerAlertView {
+    func displaySpinner(_ dismiss: Driver<Void>, bag: DisposeBag)
+}
+
 extension UIViewController: StatusAlertView {
```
jhugman: 😕 

Putting this on `UIViewController` is a little over-general. To be sure, not wrong, just… I'll keep reading.

lockbox-ios/Common/Extensions/UIViewController+.swift
```diff
@@ -4,6 +4,7 @@
 
 import UIKit
 import RxSwift
+import RxCocoa
 
 protocol StatusAlertView {
```
jhugman: Nit: naming. This shouldn't be called `View` if it isn't a `UIView`.

Suggest: `StatusAlertDisplay`? `InfiniteProgressDisplay`?

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -9,9 +9,11 @@ import RxDataSources
 import Storage
 import Shared
 
-protocol ItemListViewProtocol: class, AlertControllerView {
+protocol ItemListViewProtocol: class, AlertControllerView, SpinnerAlertView {
```
jhugman: This makes sense to have `SpinnerAlertView` here, but only if `UIViewController` isn't extended by `SpinnerAlertView`.

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -124,37 +135,26 @@ class ItemListPresenter {
...
+                        self.view?.displayFilterCancelButton()
+                    }
+                })
 
         let listDriver = self.createItemListDriver(
                 loginListObservable: itemListObservable,
                 filterTextObservable: filterTextObservable,
-                itemSortObservable: itemSortObservable
+                itemSortObservable: itemSortObservable,
+                syncStateObservable: self.dataStore.syncState
```
jhugman: Oh neat.

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -196,13 +224,46 @@ extension ItemListPresenter {
...
     }
 
+    private func placeholderComputingItemListObservable() -> Observable<[Login]> {
+        let hideSpinnerObservable = self.dataStore.syncState
+                .filter {
+                    $0 == SyncState.Synced
+                }
+                .map { _ in
+                    return ()
+                }
```
jhugman: Wut. Comment needed?

lockbox-ios/Common/Extensions/UIViewController+.swift
sashei: What if we want to throw up a spinner somewhere else? :p

sashei:
@nickbrandt I've addressed the smaller changes (sorting option disabled, no separator line on the empty state view). @changecourse and I discussed filing a follow-up issue for the visual state of the search bar, there's not a trivial way to do that unfortunately :( 

I think I got the placeholder sticking where it's supposed to be, let me know how it looks to you.

sashei:
@nickbrandt I've addressed the smaller changes (sorting option disabled, no separator line on the empty state view). @changecourse and I discussed filing a follow-up issue for the visual state of the search bar, there's not a trivial way to do that unfortunately :( 

I think I got the placeholder sticking where it's supposed to be, let me know how it looks to you.

lockbox-ios/Common/Extensions/UIViewController+.swift
jhugman: Right, I'm just seeing the other `StatusAlertView` and `AlertControllerView` on `UIViewController+.swift`.

If this was the first time you'd done this, I'd advise on an approach which `viewController.present(_ view: StatusAlertView, animated: Bool = false)`, with a convenience constructor on `StatusAlertView`.

However, this is an established pattern, so carry on.

nickbrandt:
@sashei This looks great! Thanks for fixing those items, and sounds good on the disabled state 👍

PR #354, opened 2018-05-10
Set search/filter to a darker placeholder text color

### LINKED ISSUE #317: Update the filter placeholder text color on the Entry List for better contrast
Currently, the placeholder text color for filtering the Entry List is a bit light. Let's update this to have a bit more contrast.

New color: #737373

devinreams:
Fixes #317

By definition, the UITextField placeholder text is a system provided color. This change appears to override it, but a) may not be officially supported/allowed and b) is imprecise because some additional alpha is being applied that I can't calculate.

screen shot 2018-05-10 at 12 49 17 pm

screen shot 2018-05-10 at 12 49 58 pm

### DIFF
--- a/lockbox-ios/Storyboard/Base.lproj/ItemList.storyboard
+++ b/lockbox-ios/Storyboard/Base.lproj/ItemList.storyboard
@@ -45,6 +45,11 @@
                                                     <nil key="textColor"/>
                                                     <fontDescription key="fontDescription" type="system" pointSize="15"/>
                                                     <textInputTraits key="textInputTraits" autocorrectionType="no"/>
+                                                    <userDefinedRuntimeAttributes>
+                                                        <userDefinedRuntimeAttribute type="color" keyPath="_placeholderLabel.textColor">
+                                                            <color key="value" white="0.24812825520833334" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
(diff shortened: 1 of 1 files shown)


changecourse:
@devinreams Can you confirm our Disabled state color for this field too please? Just wanting to make sure that we're not applying this change globally and adversely affecting the lighter shade.

changecourse:
That's a good question... it probably hasn't been implemented yet, which may be why you're having trouble finding it? 

It would be in two instances... No Entries due to "No Results Found" via filtering in search OR No Entries due to "No Saved Logins" via sync.... 

FWIW, our intent is to have the entire search field (text, field, icon) at 30% opacity for the disabled state.

devinreams:
>It would be in two instances... Syncing State (spinning wheel) OR No Entries due to "No Saved Logins" via sync....

@changecourse as we discussed today with @sashei, in both instances (aka really just "no entries") there is no concept of a disabled state yet. We agreed we should create a follow-up story to add that concept as it will require new components: #364 

Therefore, I believe we considered this branch "done" as it ultimately does what it should

1. change the color for the placeholder text to have more contrast
  screen shot 2018-05-14 at 12 08 12 pm

2. confirm the text input by user is still a good contrast
  screen shot 2018-05-14 at 12 08 06 pm

@changecourse @sashei Agreed? Is this good to go?

PR #350, opened 2018-05-10
copy-frameworks with sync libraries

sashei:
resolves #315

PR #356, opened 2018-05-11
Fix auto-lock timer

### LINKED ISSUE #370: "Lock Now" doesn't lock now, and neither do time-based lockouts
There is lot's of weirdness with the fingerprint lock. This maybe is #227 but im filing just in case (esp. since that issue isn't marked as P1, and this seems broken to me).  The main reproducible one I see is that if you press "lock now" then cancel, you can still access passwords. 

STR:
1.  Open lockbox (and unlock with fingerprint if necessary)
2. Press the settings gear to bring up the menu
3. Press lock now, and you are prompted for fingerprint
4. Press cancel or the home button and you are just let into the app

Time based lockout  doesn't seem to work either. I tried setting a lockout policy of one minute and then:

a. Leaving the app open for one minute
b. Backgrounding app (home button) for one minute then reopening app
c.  letting screen lock (about 2 mins) 

In all cases re-opening the app showed the users passwords without a prompt.

I wasn't sure in this last case, since unlocking the phone, might count as unlocking the device, but the cases a & b.

AH. So for the second issue (timer one) actually it looks like the timer starts when the user "kills" the app (swipe up from task manager) not from when the app is backgrounded. So maybe this was a design choice. I would  advise we revisit this decision - my hypothesis would be that users actually kill apps very infrequently (anecdotally my experience on FxOS was that users had trouble understanding the difference between the app being backgrounded (home button) and the app being closed (via the task manager). If users never "close" apps then this feature is ineffective as a security mitigation.

### LINKED ISSUE #379: Auto Lock does not activate
Steps to reproduce:

- Install lockbox, sign in, see your logins
- Background the app
- Come back later, more than 5 minutes

Expected: app asks to unlock
Actual: app opens without asking anything and shows logins

iPhone X with Face ID, iOS 11.3.1, Lockbox 1128

joeyg:
I was testing the auto lock timer and it seems to be not be firing. I thought this worked before but maybe I missed a test case.

Depends on #359 (added by @devinreams, "Welcome" screen and locking may not be firing as expected)

Resolves #370 (added by @devinreams)
Resolves #379


lockbox-ios/Store/AutoLockStore.swift
```diff
@@ -114,7 +117,9 @@ class AutoLockStore {
     }
 
     @objc private func lockApp() {
-        self.dataStoreActionHandler.invoke(.lock)
-        self.userDefaults.removeObject(forKey: SettingKey.autoLockTimerDate.rawValue)
+        DispatchQueue.main.async {
+            self.dataStoreActionHandler.invoke(.lock)
```
sashei: I can't think of a reason that the userdefaults setting or lock function would need to happen on the main thread, can you justify this? I believe you about the timer :)

devinreams:
>>I was testing the auto lock timer and it seems to be not be firing. 

>I believe you about the timer :)

Hmm yeah I think you're right. I can reproduce the timer not firing anymore. With Build 1045 at #350 I don't see it lock when in the foreground nor when backgrounded.

Thanks for catching this. It was definitely working for me at #345 _after_ @sashei made some changes. I'm looking back and can't tell what would have affected it or what I missed in testing since then. 🤔

lockbox-ios/Store/AutoLockStore.swift
joeyg: @sashei hmmm not sure. I'll double check this later. My thought was that it would need to eventually get to the main thread to update the UI to show the lock screen.

sashei: I think if there's not a reason for main-thread dispatching here, I'd rather have the listeners bind to the UI thread when they _receive_ events.

joeyg: looks like this was unnecessary anyways - thanks for calling that out. I've removed it from the PR.

joeyg:
@devinreams hmmm saw the same thing. One time I got the touch id screen to pop up but it didn't show the welcome screen. I'll have to see why the Welcome screen isn't showing up.

devinreams:
> One time I got the touch id screen to pop up but it didn't show the welcome screen.

Ah yes, we were seeing that when re-testing at #366. Something must have been introduced recently, we'll keep looking at it and see if its related/the cause.

joeyg:
nevermind - this doesn't seem to fix it either. will have to keep looking.

sashei:
hey @joeyg , before you spend too much time on this I found a really subtle bug that I'm about to push a fix for in #366, maybe check and see if that fix, fixes the autolock timer for you......

joeyg:
@devinreams thanks for making the build. I was able to get the timer to go off and lock the app when I had it set for 1 min and kept the app forgrounded. Let me know if it works for you and I’ll remove some of the hacks to see if the simple change resolved it as well

devinreams:
>I was able to get the timer to go off and lock the app when I had it set for 1 min and kept the app forgrounded.

@joeyg Same here (on iPhone X)! 🎉 

I tested with the app open in foreground on list view and device unlocked (screen on)  ✅ 

I also tested by tapping into an entry detail, device unlocked (screen on) ✅ 

I also confirmed "on app exit" locked immediately after sent to background (reopen = locked). ✅  

@mozilla-lockbox/ux could someone else please triple check and make sure this locking behavior works as expected again on another device before we merge this? 👍 👎

joeyg:
@devinreams @sashei would you like me to remove the changes to `RootPresenter` and see if it still works? That was just a guess.

sashei:
@joeyg yes, that would be good -- those calls should already be on the main thread anyway because they are coming from `Driver`s (see documentation: https://github.com/ReactiveX/RxSwift/blob/master/Documentation/Traits.md#driver)

joeyg:
Updated the PR to remove `RootPresenter` changes.

devinreams:
Re-tested latest change on iPhone X hardware and 1 minute timer.

After 54 seconds...

lock-in-forground

I also went back to the default 5 minute timer. My device locked (screen black) and after 5+ minutes I unlocked my phone and Lockbox was on the Welcome screen prompting for Face ID. ✅ 

I've hammered on this as much as I can think of for now. I believe this one is good to go. 😅

joeyg:
Thanks for the extra round of testing @devinreams. It worked for me as well.

devinreams:
>Thanks for the extra round of testing @devinreams. It worked for me as well.

All thanks to you @joeyg.

Sorry again for the run-around and rework as things re-stabilized there. 🙏

devinreams:
Sorry for prematurely merging here, I see there are still some gremlins in there. I created a follow-up issue to isolate and get these cases reproducible: #383


Thanks @st3fan for your thoughts and @nickbrandt for more testing.

PR #366, opened 2018-05-14
point to functional sync lib

### LINKED ISSUE #359: Lockbox 1.0 (1054) crashes at startup
iPhone SE (10.3)
iPhone 8 Plus (11.3)
TF build 1.0 (1054)

1. Install the build from Testflight.
2. Try to start the app.

Result: The app crashes.

After crashing a couple of times, the app opens.

[Firefox Lockbox-2018-05-14-093941.ips.beta.zip](https://github.com/mozilla-lockbox/lockbox-ios/files/1999588/Firefox.Lockbox-2018-05-14-093941.ips.beta.zip)

sashei:
resolves #359


devinreams:
Note to selves: mozilla-lockbox/lockbox-ios-fxa-sync#6 will merge so we can send to iTunes Connect. Then mozilla-lockbox/lockbox-ios-fxa-sync#7 will merge so this works.

So this PR can actually point `lockbox-ios-fxa-sync` back to "master" (or the merge commit SHA, if we prefer stick to point-in-time) in `Cartfile`.

devinreams:
>    A shell task (/usr/bin/env git checkout --quiet --force 3ac7d1247ee62f49f8064c89f122161acc236473 (launched in /Users/buddybuild/Library/Caches/org.carthage.CarthageKit/dependencies/lockbox-ios-fxa-sync)) failed with exit code 128:
62
    fatal: reference is not a tree: 3ac7d1247ee62f49f8064c89f122161acc236473

Even though the latest commit was `3ac7d1247ee62f49f8064c89f122161acc236473` and that's what Carthage sets in the `resolved` file I think it needs to be `5a51446813ef29d8adb9e3f3755ee64db9235fc5`? 😖

devinreams:
Build 1090 is looking much better over here

@mozilla-lockbox/product would someone else be able to try a new buddybuild download on other device(s) to help double check?

devinreams:
@sandysage also confirmed via Slack there was no crash 👍 

I also confirmed the app (Build 1090) successfully uploads to iTunes Connect 🍏 

---

But I did run into the first-run crash again on iPhone X. It looks a bit different than anything I've seen:

- download and launch Lockbox
- it "hangs" on the Launch screen for what seems much longer than expected
- crashes back to home screen
- but with the Face ID permissions prompt appearing over the home screen
- Allow permissions
- open Lockbox
- quickly opens to Launch then Welcome screen with "Get Started"

@sashei would you like to try and debug this or save it for a follow-up issue to isolate?

devinreams:
@sashei OK! After your last commit, I think there's some good news and some bad news here...

The **good** news: I no longer get the long pause and crash on start! 🎉 

The **bad** news: I still do get prompted for biometrics (Face ID) _immediately_ without "Get Started". So, I end up Face ID'ing then land on an empty list view (because I haven't authenticated with FxA yet).

If I "Lock Now" in this state I get thrown back to the "Welcome" screen which prompts me to "Get Started" and authenticate, no biometrics prompt .. so from there things are back on track.

But the premature biometrics prompt (hmm related to #363 perhaps?) is certainly unexpected...

devinreams:
Two things:

1. This *seemed* related to if the autolock setting/user preference was changed from default...
2. But, now I get a first run crash on `self.account.destroyDevice()`.

Regarding autolock setting: if I kept the default preference (5 minutes), then re-installs of the app did not prompt for biometrics on first run. But if I changed the preference to 1 minute, re-installs of the app _did_ prompt for biometrics on first run. Hence commit cd6b2b8.

As to the new crash: I consistently get a crash on first run even if I do a full device reset/erase settings and install the app: 

`Thread 3: EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)`
on line 344 in `FxALoginHelper.swift`
which is `self.account.destroyDevice()`
and then points to `DataStore.reset()`

sashei:
So...

I've figured out what was causing the bug that @devinreams is seeing, it's deleting the app while it's locked and reinstalling. We read the keychain `locked` value on initialization in the Datastore, so it doesn't have a change to get reset yet. I don't know how high priority fixing that is for me, but maybe others have opinions..?

devinreams:
>it's deleting the app while it's locked and reinstalling

Is there a way to get un-broken? Otherwise everyone (self included) who has the app installed and had it locked is potentially going to be permanently stuck here, right?

sashei:
@devinreams yes, you can authenticate with biometrics, see nothing in the app and then hit "Lock Now" to kick you back to the welcome screen, at which point the "locked" state will have figured itself out and you can login with FxA normally. 

This is suboptimal but fixing the datastore will be cumbersome and this seems like an edge case.

PR #368, opened 2018-05-15
fixes filtering / cancelling

### LINKED ISSUE #351: Search/filter and keyboard not quite working as expected
1. [x] Change "Cancel" behavior: to dismiss keyboard anytime (if text is input or not) and clear the search field (therefore hide "Cancel" button) in other words: Cancel button only, and always, appears when keyboard is up
2. [x] Fix "Cancel" bug: type one character and keyboard dismissed and search stops. also may result in a filtering loop that looks like an infinite/automatic scroll kinda like the matrix or something...
3. [x] Set the "Done" return key to dismiss the keyboardd
4. [x] Change "Clear" (x) behavior: to just clear the input and not dismiss the keyboard, so user can start typing again

---

From Build 1014 on iPhone X to demonstrate "Cancel" bug:

- open app, sign in, entries load into list view 🆗
- tap into search/filter input box
- type letter 'a', keyboard immediately dismissed ⚠️ 
- tap into search/filter input box
- type letter 'b', keyboard immediate dismissed ⚠️ 

Clearing is OK:

- tap into search/filter input box (with letters already input)
- tap clear (X) button 🆗 

Canceling does nothing?

- tap into search/filter input box
- type letter 'a'
- tap "Cancel" button, nothing happens ⚠️ 

I'm just guessing but it seems like the "Cancel" action is happening on every keypress? 🙃 

Related to #313 #319

sashei:
fixes #351

### DIFF
--- a/lockbox-ios/Action/ItemListDisplayAction.swift
+++ b/lockbox-ios/Action/ItemListDisplayAction.swift
@@ -14,6 +14,10 @@ enum ItemListSortingAction: ItemListDisplayAction {
     case alphabetically, recentlyUsed
 }
 
+struct ItemListFilterEditAction: ItemListDisplayAction {
+    let editing: Bool
+}
+
 class ItemListDisplayActionHandler: ActionHandler {
     static let shared = ItemListDisplayActionHandler()
     private let dispatcher: Dispatcher
--- a/lockbox-ios/Presenter/ItemListPresenter.swift
+++ b/lockbox-ios/Presenter/ItemListPresenter.swift
@@ -122,14 +122,31 @@ class ItemListPresenter {
     }()
 
     lazy private var syncPlaceholderItems = [
-        ItemSectionModel(model: 0, items: [
-            LoginListCellConfiguration.Search(
-                    cancelHidden: Observable.just(true),
-                    text: Observable.just("")
-            ),
-            LoginListCellConfiguration.ListPlaceholder
-        ])
+        ItemSectionModel(model: 0, items: self.searchItem + [LoginListCellConfiguration.ListPlaceholder])
     ]
+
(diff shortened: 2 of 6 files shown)


devinreams:
Small update: the keyboard seems to be dismissed if the input results in a re-filter of the results in the list.

For example: type "$" and it filters to 0 results, then you can re-invoke the keyboard and press any number of characters un-interrupted.

Another example: type "a" and it filters to 10 results, then re-invoke the keyboard and press "b" and it filters to 4 results, it disrupts/dismisses the keyboard. Re-invoke the keyboard and press "a" again, it filters to 1 result, it disrupts/dimisses the keyboard.. etc.

sashei:
@nickbrandt from one of the later comments on #351: 
> Change "Cancel" behavior: to just dismiss keyboard (and hide "Cancel" button), aka Cancel button only appears when keyboard is up. does not clear text field

sashei:
!!!! I can't keep track anymore, sorry Nick! I'll update the behavior.

nickbrandt:
@sashei No worries! I know there has been a lot of back and forth on this one.

devinreams:
Ugh, conflict hell. 😔 Here's hoping it's not so bad @sashei  🤞 

screen shot 2018-05-17 at 11 58 02 am

---

And sorry about the mixup on the revised "Cancel" button expectations. I obviously missed that too, even from my own description. Thanks for catching that @nickbrandt and jumping on it @sashei ⚡️

devinreams:
@nickbrandt good catch! Now with the latest change from @sashei are you seeing the "Cancel" button back where it should be?

nickbrandt:
@sashei @devinreams "Cancel" is back and everything is working as desired. Great work!

PR #375, opened 2018-05-16
only route to list from appropriate places

### LINKED ISSUE #347: Settings page is automatically dimissed when the logins list is populated
master 3ffd17f

1. Login to Lockbox.
2. Before the logins list gets populated, enter Settings and wait.

Result: The Settings page is automatically dismissed when the logins list is fully populated.

sashei:
resolves #347

### DIFF
--- a/lockbox-ios/Presenter/RootPresenter.swift
+++ b/lockbox-ios/Presenter/RootPresenter.swift
@@ -69,11 +69,8 @@ class RootPresenter {
                     $0.1
                 }
                 .subscribe(onNext: { state in
-                    switch state {
-                    case .NotSyncable:
+                    if state == .NotSyncable {
                         self.routeActionHandler.invoke(LoginRouteAction.welcome)
-                    default:
-                        self.routeActionHandler.invoke(MainRouteAction.list)
                     }
                 })
                 .disposed(by: self.disposeBag)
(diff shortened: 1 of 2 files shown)


devinreams:
>this is to stop the Settings screen being yanked away from the user after the sync has finished.

Precisely. 👍 

---

From the bug: "Before the logins list gets populated, enter Settings and wait."

Previous Result: The Settings page is automatically dismissed when the logins list is fully populated.

devinreams:
>this is to stop the Settings screen being yanked away from the user after the sync has finished.

Precisely. 👍 

>Description of in the PR would be nice. 

Certainly true, it may help speed up and focus reviews.

Here's maybe how I'd put it...

---

From the bug: "Before the logins list gets populated, enter Settings and wait."

Previous Result: The Settings page is automatically dismissed when the logins list is fully populated.

Current Result in this PR: Signing in and immediately opening the Settings page does not result in the settings page being dismissed. The list populates in the background and items are present when returning to the list view.

waiting-for-sync

devinreams:
I also tested by tapping the "Secret Refresh Button" (the list view heading) and then quickly navigating into detail item views and the settings view again to try and catch it.

I was never able to trigger a "snap back" to the list view. ✨

sashei:
@jhugman yes, routing to the list view from the FxA login screen was already in place.

PR #380, opened 2018-05-17
update titles for entries

### LINKED ISSUE #193: Provide entry titles in the app to make entries more human readable
When viewing the entry list, I want to quickly scan the entry list by title (e.g., Facebook instead of https://www.facebook.com or even www.facebook.com), so I can quickly access my username and password for logging in. 

Problem: No Entry Name (title) currently available in Saved Logins so we need to derive one

## Acceptance criteria
* Should be able to view the list of entries by the origin
* Default alphabetical ordering on the origin should make sense (eg facebook comes before twitter)
* Should not not include the protocol (http://, https://, www.)
* Should not see the suffix or anything after the suffix (.com/{*})
* Should support future goals of editing the entry name #174

## Potential solutions
1. Solution A: funky string manipulation, get domain name, drop suffix, close enough? www.chase.com = Chase, www.bankofamerica.com = Bankofamerica (Medium, not dependent)
2. Solution B: actually change the schema to include entry title (High, very dependent, long term roadmap)
3. Solution C: drop off protocol (drop https://) just show origin, just show TLD? Examples: chase.com, bankofamerica.com, but google.com could be accounts.google.com, mail.google.com (Low)

sashei:
resolves #193

### DIFF
--- a/lockbox-ios/Common/Extensions/String+.swift
+++ b/lockbox-ios/Common/Extensions/String+.swift
@@ -18,9 +18,9 @@ extension String {
 
     func titleFromHostname() -> String {
         return self
-                .replacingOccurrences(of: "http://", with: "")
-                .replacingOccurrences(of: "https://", with: "")
-                .replacingOccurrences(of: "www.", with: "")
+                .replacingOccurrences(of: "^http://", with: "", options: .regularExpression)
+                .replacingOccurrences(of: "^https://", with: "", options: .regularExpression)
+                .replacingOccurrences(of: "^www.", with: "", options: .regularExpression)
     }
 
     private func sha256(_ data: Data) -> Data? {
--- a/lockbox-ios/Presenter/ItemListPresenter.swift
+++ b/lockbox-ios/Presenter/ItemListPresenter.swift
@@ -240,7 +240,7 @@ extension ItemListPresenter {
                             .sorted { lhs, rhs -> Bool in
                                 switch latest.sortingOption {
                                 case .alphabetically:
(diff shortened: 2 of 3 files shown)


changecourse:
This generally looks good on our end, though it would be nice to also strip "www." from saved logins that have it; otherwise, all "www." entry names will just live at the bottom of the list... 

We also need to port this logic to the Entry Detail title, to be displayed in the header.

lockbox-ios/Common/Extensions/String+.swift
```diff
@@ -16,6 +16,13 @@ extension String {
         return shaData.base64URLEncodedString()
     }
 
+    func titleFromHostname() -> String {
+        return self
+                .replacingOccurrences(of: "http://", with: "")
```
linuxwolf: `replacingOccurrences` looks for anywhere within the string?

I think it would be best to limit to the start of the string.

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -219,7 +219,7 @@ extension ItemListPresenter {
                             .sorted { lhs, rhs -> Bool in
                                 switch latest.sortingOption {
                                 case .alphabetically:
-                                    return lhs.hostname < rhs.hostname
+                                    return lhs.hostname.titleFromHostname() < rhs.hostname.titleFromHostname()
```
devinreams: Awesome, that seems even easier than I had hoped. Thank you 🙇

PR #385, opened 2018-05-17
tweak list display logic so that list displays when launching in unlocked state

### LINKED ISSUE #381: Black screen after app is killed in the background
Tested on Build 1139, iPhone X, iOS 11.3.1

- Launch application
- Background application
- Kill application from the app switcher
- Start application

The app will start up with a black screen.

This may be related to my comment about using using an NSTimer for the lock at https://github.com/mozilla-lockbox/lockbox-ios/pull/356#issuecomment-389973646

Killing the app may not sound like something users do, but they ... actually do. It is also a decent representation of what happens when iOS decided to kill the app, which could happen as soon as when you switch to another app that needs resources.

sashei:
resolves #381


devinreams:
@st3fan @nickbrandt Thanks for reporting and confirming this. ❤️ 

Would you be able and willing to help re-test against the latest buddybuild on this branch (currently 1144)?

nickbrandt:
@devinreams this seems to be fixed for me, can't replicate the issue anymore.

devinreams:
Thanks @sashei! Every time I read this `bad launch` branch name I couldn't help but think:

bad-dog

Bad launch!

PR #389, opened 2018-05-17
reset auto lock timer when setting changes

### LINKED ISSUE #383: Autolock with a timer doesn't consistently work
Follow up to #356 

- [x] when set to "Never" or a different setting the app sometimes autolocks anyway or at a weird interval: make sure AutoLockStore listens for changes to the setting
- when in background the app sometimes does not lock ❓ 
- when in foreground with a timer the app sometimes does not lock❓

We need better definitions of when those "sometimes" are...

joeyg:
I would be interested to see if this resolves #383.


joeyg:
@devinreams thanks for the testing!! It’s been working well for me too

PR #390, opened 2018-05-18
Change global background container color (#367)

### LINKED ISSUE #367: Change global background container color
Change the background container of all screens with a background of #EFEFF4 to a new background of Grey20 #EDEDF0  

This was an oversight from UX regarding default system colors vs. photon colors, and has unintended consequences for color shifts on elements once we begin work on #330 (you can pull down the search row, which more drastically reveals the color differences between the two fields)

michaelgill1969:
Change viewBackground variable in color struct in Constant.swift from
#EFEFF4 to #EDEDF0.

Resolves #367


devinreams:
Thanks for your contribution, @michaelgill1969!

@mozilla-lockbox/ux I checked around the list, detail and settings views and believe this gets everything you were looking for in #367. Can you please help double check?

img_2892
img_2893

changecourse:
It would appear that our Entry Detail and Settings>Account Detail page are still pulling a different background (#EFEFF4) 

@devinreams is something overriding the global color definition in those instances, perhaps?

devinreams:
>It would appear that our Entry Detail and Settings>Account Detail page are still pulling a different background (#EFEFF4)

>@devinreams is something overriding the global color definition in those instances, perhaps?

Yes. I believe storyboards set their own backgrounds with hard-coded colors (not using our nice constants). I'm not quite sure how we can add code to override those to use use our nice constants. 

For example, here's where the "Account Detail" color is set that you end up seeing:

https://github.com/mozilla-lockbox/lockbox-ios/blob/e6f3a00c203a4796c831fec1050fb99a3fea255f/lockbox-ios/Storyboard/Base.lproj/AccountSetting.storyboard#L64

michaelgill1969:
I have it working so it tests correctly on my Simulator, but I have a question.  Would you prefer that the color be set:

1. in the storyboard (tested - works)
2. programmatically in each view subclass individually (tested - works)
3. programmatically in an extension to UIView (untested - working on it)

devinreams:
That's so much clearer and maintainable, @michaelgill1969.

Thank you again for your contribution. 🥇

devinreams:
Thanks again for contribution number two, @michaelgill1969! 🥇

Issue #328, opened 2018-05-03
Handle FxA email confirmation and quitting the app while logging in


sashei:
Case to handle:

- User does email / password via FxA & gets email prompt screen
- User hard quits lockbox app, losing that screen
- User confirms email & returns to lockbox app

devinreams:
Note to selves: this should also handle the scenario / bug at #360 where no force quit happens but a user simply drops from sign in form into the app (and doesn't realize there is an email confirmation waiting/expected).

changecourse:
My hope for this issue would be that we also account for always having the placeholder image and content on an empty state of the entry list... see the problem I encountered below on a previous build...

>I did encounter something strange where the entry list container was completely empty for approx 1-2 seconds, and then the empty state placeholder and text 'popped' onto the screen, with the sync overlay... though this could have to do with a delay caused by the email confirmation... our assumption here is that even if the email confirmation is in a pending state, we should still have the placeholder image and content immediately upon first load of the entry list, correct?

PR #395, opened 2018-05-18
no entries when synced placeholder

### LINKED ISSUE #44: First run without data
Opening this issue to identify the possible paths for downloading the Lockbox app without associated date:

- if there is no Lockbox data (Lockbox account/bucket assigned to this FxA) there is a path to get that set up in the extension
- if there are no entries (Lockbox is assigned to FxA but there are no entries) there is a path to add entries in the webextension

sashei:
resolves #44


devinreams:
To consider from #391: the A-Z button dissapears when tap to refresh/sync

devinreams:
Thanks for UX review. ✅ 

From standup: requested code review from @jhugman. Then we're ready to merge...

devinreams:
>I noticed the background on the text in the middle has a background different than the container, but this will be resolved once PR #390 is merged in.

That change is merged in and looks good here (confirmed #EDEDF0 throughout the background and around the "No entries found" text):

screen shot 2018-05-21 at 11 53 01 am


Final recap: code review needed then we're all set. Nice work @sashei! 🚢

changecourse:
@devinreams I did encounter something strange where the entry list container was completely empty for approx 1-2 seconds, and then the empty state placeholder and text 'popped' onto the screen, with the sync overlay... though this could have to do with a delay caused by the email confirmation... our assumption here is that even if the email confirmation is in a pending state, we should still have the placeholder image and content immediately upon first load of the entry list, correct?

sashei:
@changecourse there might be some lag with confirming the "verified" email confirmation state even when they don't require the clicky email flow. I think that adding the placeholder pieces for that state is dependent on @jhugman 's email confirmation work regardless; might be worth just noting this case in that story.

changecourse:
Understood... will bring this concern to #328 then. Thanks @sashei

devinreams:
yes

PR #400, opened 2018-05-18
check that script messages come from original url

### LINKED ISSUE #371: FxA Origin not checked in FxAView.swift message handler
In the lockbox-ios-fxa-sync repo, there is a [check](https://github.com/mozilla-lockbox/lockbox-ios-fxa-sync/blob/c4a50380c1d126074c316e8cdd4bdc0d4e4fabec/FxAUtils/FxAViewController.swift#L142) to make sure that messages that are sent from accounts.firefox.com are actually from this origin.  However in FxAView we've [commented this check ]( https://github.com/mozilla-lockbox/lockbox-ios/blob/master/lockbox-ios/View/FxAView.swift#L104) out as far as I can tell. Now it's an unlikely attack vector, but this check would seem to be important to me to ensure the integrity of the sign-in process. Is there a reason why this check has been removed?

From testing with a debugger attached, there doesn't seem to be anything stopping the WkWebView which is showing accounts.firefox.com from being navigated, and then another webpage calling _webkit.messageHandlers.accountsCommandHandler.postMessage(...)_ directly. I'm not sure what the risk is here, but it seems odd to remove the check.

Related to #351

sashei:
resolves #371


lockbox-ios/View/FxAView.swift
```diff
@@ -101,11 +103,11 @@ extension FxAView: WKScriptMessageHandler {
         // message is the same as the origin of the URL we initially loaded in this web view.
         // Note that this exploit wouldn't be possible if we were using WebChannels; see
         // https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/WebChannel.jsm
-        //   let origin = message.frameInfo.securityOrigin
-        //   guard origin.`protocol` == url.scheme && origin.host == url.host && origin.port == (url.port ?? 0) else {
-        //   print("Ignoring message - \(origin) does not match expected origin \(url.origin)")
-        //   return
-        //  }
+        let origin = message.frameInfo.securityOrigin
+        guard origin.`protocol` == self.url?.scheme && origin.host == self.url?.host && origin.port == (self.url?.port ?? 0) else { // swiftlint:disable:this line_length
```
jhugman: I think you can break this up with 

```swift
guard let url = self.url,
    origin.`protocol` == url.scheme,
    origin.host == url.host,
    origin.port == url.port else {
}
```

lockbox-ios/View/FxAView.swift
```diff
@@ -35,6 +36,7 @@ class FxAView: UIViewController, FxAViewProtocol, WKNavigationDelegate {
 
     func loadRequest(_ urlRequest: URLRequest) {
         self.webView.load(urlRequest)
+        self.url = urlRequest.url
```
jhugman: _Might_ be worth checking FxA's intention about redirects.

In fact: we should likely:

 a. Set `url` as `URL!` — because it's only going to be set once, and before it's needed.
 b. Check that navigation has finished, and iff `url` is unset, the set it — so `url` only contains the final destination of the first requested page.

sashei: I don't totally follow what you mean with b). Where should we check that the the navigation is finished? I think `loadRequest` only gets called when the view is loaded initially.

lockbox-ios/View/FxAView.swift
```diff
@@ -104,8 +104,10 @@ extension FxAView: WKScriptMessageHandler {
         // Note that this exploit wouldn't be possible if we were using WebChannels; see
         // https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/WebChannel.jsm
         let origin = message.frameInfo.securityOrigin
-        guard origin.`protocol` == self.url?.scheme && origin.host == self.url?.host && origin.port == (self.url?.port ?? 0) else { // swiftlint:disable:this line_length
-            print("Ignoring message - \(origin) does not match expected origin \(self.url?.origin)")
+        guard origin.`protocol` == self.url.scheme,
```
jhugman: I woke up last night screaming that this shouldn't be a direct deref.

You can collect the URL when the webView finishes loading.

There's a method in `WKNavigationDelegate` called 

```swift
func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!)
```

PR #402, opened 2018-05-20
pull to refresh

### LINKED ISSUE #330: Pull to refresh
from both the empty state and the entries list

1. pull down partially does not sync but provides a visual indicator
2. pull down fully (a "full" pull) does then initiate sync
3. in both cases, there is a 'snap back' interaction expected here
4. after "full" pull, the spinning/loading indicator stays in place until sync is complete
5. after sync is complete, the indicator/table collapses back to just search bar + list
6. after sync is complete, newly sync'd items appear at top of table with visual highlight (nice to have)

Related: #367 (background color)

joeyg:
Resolves #330. There was a lot of discussion on the ticket so let me know if this is missing part of the acceptance criteria. It also stores the state in the presenter but I was unsure if that was ok in this case.


devinreams:
This is really nice, @joeyg! This worked great for me in my brief testing (I removed a few logins from the test account and watched them 'disappear' here). I'll defer to @mozilla-lockbox/mobile-engineering for code review and @mozilla-lockbox/ux for any other design or interaction considerations.

A couple of open questions, perhaps to consider addressing separately:

1. The loading indicator appears to slides over top of the table view / filter input at the very last moment. After you pull "enough" to toggle the refresh, the indicator stays locked/anchored above the table but it seems once it's completed and everything collapses away, it's allowed to slide up and over the table for a brief moment. Should that fade out faster than the table scrolls up so they never touch/overlap? Should the animation be changed here somehow? 🙃 

2. The "tap the title in the nav bar" aka "secret refresh button" still exists right now. Should both exist alongside each other or should this pull to refresh behavior supersede that tap interaction (thus it should be removed)?

Thank you for yet another contribution 😍

joeyg:
Updated PR to remove the manual refresh button

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -44,6 +45,8 @@ class ItemListPresenter {
     private var itemListDisplayStore: ItemListDisplayStore
     private var disposeBag = DisposeBag()
 
+    var manualSync = false
```
sashei: I don't love this stateful way of keeping track of this piece, I'd rather see the manual sync piece be another `ItemListDisplayAction` style configuration so that we can put telemetry on it if we eventually want to do that.

joeyg:
The tests will need to be fixed but wanted to get @sashei your thoughts on the architecture. I think it probably could be cleaned up a bit.

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -208,13 +210,35 @@ class ItemListPresenter {
...
+                let isSyncing = latest.0
+                let isManualSync = latest.1.refreshing
+                return isSyncing && isManualSync
+        }.bind(to: view.pullToRefreshActive)
+            .disposed(by: self.disposeBag)
+
+        self.dataStore.syncState.filter { (syncState: SyncState) -> Bool in
+                return syncState == .Synced
+            }.subscribe(onNext: { (_) in
+                self.itemListDisplayActionHandler.invoke(PullToRefreshAction(refreshing: false))
```
sashei: This could be done in the `do(onNext:{})` operator for the previous call. You've already set up the `syncingObserver`, so even if you were going to keep this separation (I think there's an argument for separating that kind of logic tbh) I would use `syncingObserver` rather than re-implementing the filter.

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -208,13 +210,35 @@ class ItemListPresenter {
...
-        guard let view = self.view,
-              let sortButtonObserver = view.sortingButtonEnabled else {
-            return
+        guard let view = self.view else { return }
+
+        let syncingObserver = self.dataStore.syncState.map {(syncState: SyncState) -> Bool in
+            return syncState == .Syncing
         }
 
+        let isManualRefreshObserver = self.itemListDisplayStore.listDisplay
```
sashei: This is an `Observable`, not an `Observer` :p

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -224,10 +248,15 @@ class ItemListPresenter {
                 .map { _ in return () }
                 .asDriver(onErrorJustReturn: ())
 
-        self.dataStore.syncState
-                .asDriver(onErrorJustReturn: SyncState.Synced)
+        Observable.combineLatest(self.dataStore.syncState, self.itemListDisplayStore.listDisplay)
```
sashei: Similarly, you can use `isManualRefreshObserver` which you've already filtered for previously to avoid the `if let` in the drive section.

lockbox-ios/View/ItemListView.swift
```diff
@@ -91,6 +91,10 @@ extension ItemListView: ItemListViewProtocol {
         return self.tableView.rx.isScrollEnabled.asObserver()
     }
 
+    var pullToRefreshActive: AnyObserver<Bool> {
+        return self.tableView.refreshControl!.rx.isRefreshing.asObserver()
```
sashei: Rather than force-casting, make the `AnyObserver<Bool>` optional and handle that case in the presenter.

lockbox-ios/Presenter/ItemListPresenter.swift
joeyg: This one is checking for the `Synced` state in order to dismiss the pull to action. `syncingObserver` is for when it is engaged in the syncing process.

lockbox-ios/Presenter/ItemListPresenter.swift
joeyg: oops :)

lockbox-ios/Presenter/ItemListPresenter.swift
sashei: ...oops, you're right !!! I can't read ;)

devinreams:
@joeyg This is great! Thanks again for building this feature and addressing all our comments. ⚡️ 🏅 

I tested on-device (iPhone X, 1328) and the refresh behavior worked well for me. And the spinner did stay anchored / sticky above the search/filter box. 👍 

I then merged in `master` (really poorly..) and re-tested everything end-to-end (build 1334, "from Zurich") and things are still working great here. Confirmation/preparing works, list view works, pull to refresh works. Hooray! ✅ :shipit: 

---

Two notes, not of real concern but in case anyone else sees anything similar:

1. I _think_ I could get it so the spinner would only briefly stay anchored above the filter _then_ immediately drop down, while still spinning, over the filter box but after ~20 pulls it never happened again.

2. I also think I got into some weird state where pull to refresh didn't show the spinner icon but when I closed/re-opened the app everything came back working fine. I couldn't make it happen again (un)fortunately so will continue to see if I can figure out how it happened.

PR #406, opened 2018-05-21
Fix some buildtime warnings

devinreams:
I recently spotted some of these in Xcode and figured if these are safe to fix then may as well clean them up. If better solutions exist then I'm open, just wanted to get these a bit more tidy so we didn't inadvertently start ignoring warnings...

@jhugman @sashei I poked around about the "Implicit import of bridging header... is deprecated" warning and wasn't sure how/what applied to our project and how to remedy it. In case you knew, thought it was a good idea to remove, and wanted to clear it on this branch while we're at it...

### DIFF
--- a/Lockbox.xcodeproj/project.pbxproj
+++ b/Lockbox.xcodeproj/project.pbxproj
@@ -837,7 +837,7 @@
 			attributes = {
 				LastSwiftUpdateCheck = 0900;
 				LastUpgradeCheck = 0930;
-				ORGANIZATIONNAME = "Sasha Heinen";
+				ORGANIZATIONNAME = Mozilla;
 				TargetAttributes = {
 					7D1B1AC31F99028D00C1F5FF = {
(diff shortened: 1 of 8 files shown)


Lockbox.xcodeproj/project.pbxproj
```diff
@@ -837,7 +837,7 @@
 			attributes = {
 				LastSwiftUpdateCheck = 0900;
 				LastUpgradeCheck = 0930;
-				ORGANIZATIONNAME = "Sasha Heinen";
+				ORGANIZATIONNAME = Mozilla;
```
sashei: 😆

devinreams:
I've made a follow-up issue for the bridging header warnings since that seems a little more involved. Maybe, maybe not. https://github.com/mozilla-lockbox/lockbox-ios/issues/410

Either way I'll merge this where it is so we're down to 3 warnings...

PR #407, opened 2018-05-21
warn users before attempting to use app without passcode

### LINKED ISSUE #338: If device pin/biometrics not set, provide instructions to do so
With changes from issue #314, lock capabilities  are unusable in the settings screen when users do not have any available device authentication methods.

We need to communicate and support users to set this up without requiring device authentication to use the app.

## Requirements

- display the passcode prompt (when appropriate) on `Get Started` button tap in the welcome page
- if the user taps `Skip`, just route to FxA
- if the user hits set passcode, route to the settings page and remain on the welcome screen in the application
- don't bother storing the `Skip` preference, just launch it (when warranted) every time the user hits `Get Started`

## Visual Design
**InVision**
https://mozilla.invisionapp.com/share/7JEUG47ZU#/296414673_No_PIN

**Zeplin**
https://zpl.io/V4JW1Jz

Note: "Set Passcode" button would direct user to Touch ID & Passcode section of Settings app.

See also #339

sashei:
resolves #338

### DIFF
--- a/lockbox-ios/Action/ExternalLinkAction.swift
+++ b/lockbox-ios/Action/ExternalLinkAction.swift
@@ -9,7 +9,7 @@ import UIKit
 protocol LinkAction: Action {}
 
 struct ExternalLinkAction: LinkAction {
-    let url: String
+    let baseURLString: String
 }
 
 enum SettingLinkAction: LinkAction {
@@ -24,7 +24,7 @@ enum SettingLinkAction: LinkAction {
 
 extension ExternalLinkAction: Equatable {
     static func ==(lhs: ExternalLinkAction, rhs: ExternalLinkAction) -> Bool {
-        return lhs.url == rhs.url
+        return lhs.baseURLString == rhs.baseURLString
     }
 }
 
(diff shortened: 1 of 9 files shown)


devinreams:
I may be missing something but I don't believe there's a way to simulate passcode presence or not (just biometrics) on Simulator and removing a passcode from a live iPhone results in a lot of hassle and mess: so we'Il especially need extra help from folks with willing 'test' devices on.

sashei:
yeah I could not find a good way to test this in the simulator... physical device required.

lockbox-ios/Action/ExternalLinkAction.swift
```diff
@@ -6,18 +6,30 @@ import Foundation
 import RxSwift
 import UIKit
 
-struct ExternalLinkAction: Action {
+protocol LinkAction: Action {}
+
+struct ExternalLinkAction: LinkAction {
     let url: String
```
jhugman: Nit: use `let url: URL`.

lockbox-ios/Action/ExternalLinkAction.swift
```diff
@@ -32,8 +44,12 @@ class ExternalLinkActionHandler: ActionHandler {
...
-        self.openUrl(string: action.url)
+    func invoke(_ action: LinkAction) {
+        if let externalLink = action as? ExternalLinkAction {
+            self.openUrl(string: externalLink.url)
+        } else if let settingLink = action as? SettingLinkAction {
+            self.openSettings(settingLink)
+        }
     }
 
     private func openUrl(string url: String) {
```
jhugman: Nit: use `open(url: URL)`

PR #409, opened 2018-05-21
Update navigation bar styles to be consistent

### LINKED ISSUE #251: Navigation bar action link updates
In order to have consistency across the app, we want to update some of the press states for action links/icons within the navigation bar.

**Press state value: #ffffff @ 60% opacity (both text & icons)**

- [x] "Cancel" action link in FxA authentication screens + update text to "Close"
- [x] Settings icon on Entry List screen
- [x] Down caret icon (in sort) on Entry List screen
- [x] Back chevron icon on Entry Detail screen
- [x] "< Settings" action link (icon and text) on Account screen (see #337)
- [x] update padding between "<" & "Settings" (see #208)
- [x] Center the "Firefox Lockbox" text on small screen (iPhone SE #372)

Related to #223

devinreams:
Resolves #251

This may be a ton of re-use but some of it seems to be necessarily close, but not exactly the same depending on the view's needs. I may be mistaken.

That said, this should get everything into a more consistent state.

FxA sign in webview:
1-fxa-close

Selected/Highlighed A-Z sorter:
2-az-selected

Settings nav button selected (on item list):
3-settings-selected

Back button (arrow and title) selected (on item detail):
4-back-item-detail

Settings button (arrow and title) selected (on Account settings):
5-account-settings-left

FWIW the "< Settings" on the web views (FAQ, Send Feedback) is slightly different as it's not heavily customized. This gives it the standard highlighted state and pushes it off to the left a bit more than the right-shift we have on these buttons ☝️ 

Depends on #402 because I want to wait and see how the "Your Lockbox" title bar gets re-implemented (sans "refresh button") and make sure I build/test/update against that work.


lockbox-ios/View/AccountSettingView.swift
```diff
@@ -70,16 +70,22 @@ extension AccountSettingView: UIGestureRecognizerDelegate {
...
 
+        leftButton.contentHorizontalAlignment = .left
+        leftButton.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
         leftButton.titleEdgeInsets = UIEdgeInsets(top: 0, left: 5, bottom: 0, right: -20)
+        leftButton.setTitleColor(.white, for: .normal)
+        leftButton.setTitleColor(UIColor(white: 1.0, alpha: 0.6), for: .highlighted)
+        leftButton.setTitleColor(UIColor(white: 1.0, alpha: 0.6), for: .selected)
+        leftButton.setTitleColor(UIColor(white: 1.0, alpha: 0.6), for: .disabled)
+        leftButton.tintColor = .white
+        leftButton.translatesAutoresizingMaskIntoConstraints = false
```
devinreams: Whoops, I think this unnecessary addition was the problem. After removing:

screen shot 2018-05-21 at 4 15 32 pm

screen shot 2018-05-21 at 4 17 20 pm

devinreams:
It looks like the width assigned to the sorting button is what's pushing the title off to the side on smaller width screens. I'm going to shrink it down here and call this done. ✅ 

I tested it on this branch here and also @joeyg's where he changes the title from a (refresh) button back to just a title at #402.

devinreams:
screen shot 2018-05-21 at 4 54 34 pm

devinreams:
I'm likely going to close this in favor of #413 😍

PR #418, opened 2018-05-22
restart timer whenever user interacts with application

### LINKED ISSUE #405: The app auto-locks while using it
Lockbox 1.0 (1189)
iPhone 8 Plus (11.3)

1. Set the Auto Lock timer to 1 min.
2. Interact with the app for one min (navigate through the screens, filter logins, etc...)

Expected: The app doesn't lock while using it.  Should lock only after idling for 1 minute.

Actual: The app auto-locks while using it.

sashei:
resolves #405

### DIFF
--- a/lockbox-ios/Store/AutoLockStore.swift
+++ b/lockbox-ios/Store/AutoLockStore.swift
@@ -42,10 +42,20 @@ class AutoLockStore {
                 .subscribe(self.lifecycleAction)
                 .disposed(by: self.disposeBag)
 
-        self.userDefaults.onAutoLockTime.subscribe(onNext: { _ in
-            self.stopTimer()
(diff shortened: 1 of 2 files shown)


sashei:
@sandysage would you like to confirm the behavior here?

sandysage:
Shwew! Devin did some thorough testing.

I verified, with autolock set to 1 minute:
* open app and authenticate to view list view 
* let the phone rest idle for 60 seconds
   * app locks as expected

* actively use app for 5 minutes (alternating between Lockbox and another app testing copy/paste)
   * app did not lock during interactions as expected

* background Lockbox app
* actively use iphone for other apps for 60 seconds
   * return to Lockbox and app is locked as expected

This behaves as I would expect.

PR #419, opened 2018-05-22
unlock automatically when device auth has become unavailable

### LINKED ISSUE #339: Automatically unlock when no device authentication available
There is the edge case when a user locks the app, and then removes all enrolled biometrics and passcodes from the device. It seems unlikely and rare and we're acknowledging that we're just going to unlock the app without any authentication.

Depends on #338

sashei:
resolves #339


changecourse:
Agreed... I think in a future issue we may want to consider the implications of any form of auto-lock while on an unlocked device... but that's a bigger discussion.

sashei:
you know I feel like we'd talked about making all of the autolock settings unavailable when there are no biometrics, should we just do that?

changecourse:
Yes!

sashei:
@changecourse @nickbrandt what about the `Lock Now` button?

changecourse:
Probably remove as well? 

_*With the caveat*_ that we may need to revisit this once we have *OAuth* in place, to allow FxA-based locking based on using that as an authentication mechanism.

sashei:
Wait, are we saying to _remove_ the autolock setting entirely? Or just grey it out (like a preferred browser that you don't have installed)?

changecourse:
Ah, fair... we can disable, however, when we disable something, I'm inclined to provide description content to support what it would take to enable it, so it would be nice to have some sort of description around enabling a passcode in order to lock Lockbox.

nickbrandt:
@sashei makes sense to me, I'll go ahead and create a follow-up issue to handle the notification dialog for once a user disables PIN/biometrics.

PR #436, opened 2018-05-24
Login route events

irrationalagent:
I added some events for the login route: `welcome` `fxa` and `learnMore`
I need to update the metrics doc as well, I'll do that here.


irrationalagent:
I had to modify a test in `TelemetryStoreSpec.swift` that was expecting no events to be logged for `LoginRouteAction.welcome` which isn't true anymore, lmk if i done right

lockbox-iosTests/TelemetryStoreSpec.swift
```diff
@@ -58,7 +58,7 @@ class TelemetryStoreSpec: QuickSpec {
                 }
 
                 it("does not pass through non-ItemDetailDisplayActions") {
-                    self.dispatcher.fakeRegistration.onNext(LoginRouteAction.welcome)
+                    self.dispatcher.fakeRegistration.onNext(UserInfoAction.load)
```
sashei: 👍

PR #438, opened 2018-05-24
Add UI for new preparing state

### LINKED ISSUE #417: Add UI for when user is in "confirmation email" state
Follow up to #328 
Requires #412

## Visual Design
**InVision**
https://mozilla.invisionapp.com/share/7JEUG47ZU#/298634393_Entry_List_-_Confirm_Email

**Zeplin**
https://zpl.io/2y7o0wn

jhugman:
Fixes #417.

This PR adds a new empty state placeholder for when the user has to verify the account by confirming their email address.


nickbrandt:
@jhugman looks like some tests are failing here, so we're unable to install the build from buddybuild :(

sashei:
@nickbrandt hey nick, I'm picking up this PR from here, it's not completely done yet so that is why you can't install.

devinreams:
Thinking out loud: I think the best way for us to try and trigger an FxA "confirmation required" state might be to simulate a different location. Xcode simulator can do that. A VPN can as well. 🤔

jhugman:
This is now ready for review /cc @sashei 

If you want to force a confirmation dialogue, I have found toggling my VPN to a different location to work well, or using a different account.

Please also pay attention to differences between "Disconnect this account" and "Fresh install" flow — there shouldn't be any. 

Also: known issue: a brief flash of email confirmation screen when confirmation isn't required: this is likely a corner case — because most users will definitely need to confirm their account on first time login.

devinreams:
>when_the_datastore_is_locked__regardless_of_synced_state__routes_to_the_welcome_view

@jhugman @sashei it looks like the same 3 tests are still failing from before. is that still expected?

lockbox-ios/Store/DataStore.swift
```diff
@@ -181,8 +181,7 @@ class DataStore {
                 })
                 .disposed(by: self.disposeBag)
 
-        self.setInitialSyncState()
-        self.setInitialLockState()
+        self.setInitialState()
```
sashei: 😍

devinreams:
Thanks for fixing those tests, yo. ✅ :shipit: 🚀 

I re-tested this on-device (iPhone X, 1330) with VPN trickery (Sweden) and couldn't seemingly get anything to break.

I'd like to merge this as-is, catch any UX/UI feedback early next week so: PI and our testers can hammer on it (and we can build against it). Please direct any support requests directly to me. 🙃

PR #457, opened 2018-05-30
Remove onAppExit autolock setting for initial release

### LINKED ISSUE #456: Remove 'on exit' from autolock options
Right now the app display does not retain it's state once the app locks. This means search (#398) & sort (#425) does not persist, nor does an entry detail view (#408).

This is most problematic for those who have their autolock timer set to 'on exit'. Every time the app is sent to the background it locks. If the user is in the middle of a task, such as finding a desired list item or copying credentials from a detail view, they are interrupted mid-task when the app locks.

To keep the autolock option of 'on exit' we need to prioritize more work to make that a much less painful experience.

As such, let's remove this setting. At least for now.

If we need to add 'on exit' back to our autolock options, I need to better understand the desired use case. What benefit does this option provide that isn't achieved with the 1 minute lock option?

devinreams:
Fixes #456

@sashei sorry for my ignorance: is this `lifecycleAction` in `AutoLockStore.swift` still needed now?

https://github.com/mozilla-lockbox/lockbox-ios/compare/456-remove-on-app-exit?expand=1#diff-983f28423e9d2c3226bc33cfdb2ce20cR66

### DIFF
--- a/lockbox-ios/Store/AutoLockStore.swift
+++ b/lockbox-ios/Store/AutoLockStore.swift
@@ -59,14 +59,7 @@ class AutoLockStore {
                 .disposed(by: self.disposeBag)
     }
 
-    private func lifecycleAction(evt: Event<LifecycleAction>) {
-        if evt.element == .background {
-            self.userDefaults.onAutoLockTime
-                    .take(1)
-                    .subscribe(onNext: { (latest: AutoLockSetting) in })
-                    .disposed(by: self.disposeBag)
-        }
-    }
+    private func lifecycleAction(evt: Event<LifecycleAction>) {}
 
     private func resetTimer() {
         self.stopTimer()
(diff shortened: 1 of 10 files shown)


lockbox-ios/Store/AutoLockStore.swift
```diff
@@ -63,11 +63,7 @@ class AutoLockStore {
         if evt.element == .background {
```
sashei: I would remove this conditional entirely rather than keeping the pointless subscription.

devinreams: @sashei Ah ok, so if I understand you correctly: remove all of (currently) lines 63-72 and leave it empty like this?

```private func lifecycleAction(evt: Event<LifecycleAction>) {}```

File: lockbox-iosTests/ItemListPresenterSpec.swift [hunk 2]
```diff
@@ -264,7 +264,7 @@ class ItemListPresenterSpec: QuickSpec {
                             }
                         }
                     }
-
```
sashei: 👾

PR #472, opened 2018-06-05
onboarding confirmation screen

### LINKED ISSUE #325: Onboarding confirmation with security theatre (onboarding)
When I’ve successfully authenticated with FxA, I want to know that my logins have been added to my Lockbox and that I am ready to use the app

## Acceptance criteria 
* Should showcase the security of the product
* Should include an option to learn more about the security which deep links to an FAQ
* Should confirm that a user is finished with the first-run setup and ready to launch their Lockbox.
* Should be the final step as part of the first-run onboarding experience.

## Visual Design
**InVision**
https://mozilla.invisionapp.com/share/7JEUG47ZU#/294915219_Security_Theater

**Zeplin**
https://zpl.io/VkmvpLK

sashei:
resolves #325


sashei:
@changecourse yessss, I forgot to document the issue you're having with the screen dismissing after 3 seconds. This happens when the user doesn't need to confirm their email address; the app starts syncing and goes straight to the list when that happens. This shouldn't be an issue when we switch to OAuth, and seems like an edge case when most users will have to go through the email confirmation step on first login anyway.

I'll change the title and look into the disconnect behavior. thanks !!

devinreams:
*cruising through to check on and recap the status of PRs*

>>I'm encountering an issue where the screen is dismissing automatically, after about 3 seconds or so

>seems like an edge case when most users will have to go through the email confirmation step

@sashei Interesting.. I feel like this is worth adding to a "known issues" or a similar list. Or should we just hold off until after next week and we (hopefully) swap out with OAuth?

>>I'm also encountering this screen as a brief overlay when disconnecting and reconnecting my account

>I'll change the title and look into the disconnect behavior.

@sashei I assume you're still looking into the disconnect behavior from @changecourse?

devinreams:
Yay! Thanks @sashei. This should make for a nice "first run" experience. ❤️

PR #473, opened 2018-06-05
added accessibility for filter button

### LINKED ISSUE #465: Make the Cancel button in the filter field accessible
currently, when using a screen reader, the `Cancel` button is unselectable even after it appears to a sighted user typing text into the filter field

### LINKED ISSUE #469: "search your entries" is not tappable at the magnifying glass icon and all the way to right edge
Follow up from #441

sashei:
resolves #465 
resolves #469


devinreams:
>make search icon start search 056fa81 

@sashei That's a nice improvement, thank you.

I also just re-ran through the VoiceOver experience with the "actions available" and its working great. 👍 

I consider both issues fully resolved now. @mozilla-lockbox/ux do you want to test this out as well?

PR #474, opened 2018-06-06
show syncing notification while ready as well as syncing

### LINKED ISSUE #444: Syncing entries notification flashes too quickly
When I log in, first run, to Lockbox I see my list of entries appear at the same time as the loading indicator. It's really fast. Too fast to read.  Oddly enough the included gif is slowed down from the experience I had on the phone.

But then when I pull to refresh the loading indicator happens so quickly it appears as a flash.

screenrecording

sashei:
resolves #444


sashei:
I'm going to address all of these syncing notification issues in one place (see #478)

PR #475, opened 2018-06-06
make the settings button larger

### LINKED ISSUE #466: Settings icon is difficult to tap at times
This is something that we have noticed on our test device, and also came up during our pilot evaluative test with the participants device. Clicking on the Settings icon from the Entry List screen sometimes doesn't invoke the Settings screen. We are curious if this has to do with the tap target only including the icon itself and not the area around it.

Ideally, we'd want an entire bounding box to be tappable around the icon, as a square that covers the height of the header bar (and corresponding equidistant width)

sashei:
resolves #466

### DIFF
--- a/lockbox-ios/View/ItemListView.swift
+++ b/lockbox-ios/View/ItemListView.swift
@@ -301,6 +301,7 @@ extension ItemListView {
         button.setImage(tintedPrefImage, for: .selected)
         button.setImage(tintedPrefImage, for: .highlighted)
         button.setImage(tintedPrefImage, for: .disabled)
+        button.contentEdgeInsets = UIEdgeInsets(top: 10.0, left: 10.0, bottom: 10.0, right: 0.0)
         button.tintColor = .white
         button.translatesAutoresizingMaskIntoConstraints = false
         return button


sashei:
@devinreams increasing the image edge insets on the right side pushes the icon left. I could look into changing the default padding from the edge on the UIBarButtonItem if you or @changecourse / @nickbrandt  feel this would make a significant UX difference

changecourse:
@sashei It would be nice to have consistent padding equidistant from the center of the icon itself... if it's not too much trouble.

devinreams:
@changecourse Sure, one point being: adding more tap space on the right then _also_ requires shifting the button to the right an equal amount (outside of the normal bounds).

inbox_ _mozilla__10_messages__1_unread_

@sashei says its not an problem to do, my point is more to highlight that it could be something to consider (is placement okay across all screen sizes including small and large) and maintain (if Apple changes spacing we may need to update it here too). It's kinda in the same vein of how our "<Settings" and "<Back" buttons now differ a bit from the standard implementations...

changecourse:
Hmmmmmm, my aim is not to intentionally deviate from the standard implementations... 

As it was conceived and designed visually, the Settings icon should be a 20px by 20px icon within a larger (40-45px bounding box) This is in alignment to the HIG for tap sizes, and my understanding for how any clickable icon or otherwise small tap target should work within the header bar.

screen shot 2018-06-07 at 12 59 25 pm

sashei:
Ok, moving the button over involves a lot of nasty custom code -- see https://forums.developer.apple.com/thread/80075 and https://stackoverflow.com/questions/44677018/ios-11-uinavigationbar-bar-button-items-alignment/46660888#46660888

I don't really like any of these; I think the provided solution makes the button easier to tap and most people will not see the asymmetric bounding box.

changecourse:
Ah, sounds good. Thanks!

sashei:
*shakes fist* UIKit.....

PR #478, opened 2018-06-06
spinner HUD improvements

### LINKED ISSUE #444: Syncing entries notification flashes too quickly

### LINKED ISSUE #464: “Syncing Entries..” alert flashes briefly when background syncs work
### desired behavior:
do show HUD `Syncing your entries` spinner:
- after email confirmation (first sync)
- on unlock
- on app launch from background

not show during:
- background sync while app is open
- pull to refresh

HUD spinner is never displayed for less than 1 second (pending design approval on this detail), even when attendant sync completes earlier.

original issue:
~~Two observations:~~

~~1. While using the app after the “pull to refresh” #330 landed, it seems the “syncing entries” alert shows a bit too late in the sync process. Instead of when working on first run, it doesn’t show until the split second right before the items load in.~~

~~2. Sometimes a background sync causes the alert to “flash” just for a split second.~~

~~In both cases its unreadable and a bit jarring to the user.~~

sashei:
resolves #464 
resolves #444 

rxswift 😍

### DIFF
--- a/lockbox-ios/Presenter/ItemListPresenter.swift
+++ b/lockbox-ios/Presenter/ItemListPresenter.swift
@@ -241,7 +241,6 @@ class ItemListPresenter {
                 settingButtonObserver: settingButtonObserver
         )
 
-        self.itemListDisplayActionHandler.invoke(ItemListSortingAction.alphabetically)
         self.itemListDisplayActionHandler.invoke(ItemListFilterAction(filteringText: ""))
         self.itemListDisplayActionHandler.invoke(PullToRefreshAction(refreshing: false))
     }
@@ -380,7 +379,7 @@ extension ItemListPresenter {
 
     fileprivate func setupButtonBehavior(
             view: ItemListViewProtocol,
-            itemSortObservable: Observable<ItemListSortingAction>,
+            itemSortObservable: Observable<ItemListSortSetting>,
             sortButtonObserver: AnyObserver<Bool>,
             settingButtonObserver: AnyObserver<Bool>) {
         let itemSortTextDriver = itemSortObservable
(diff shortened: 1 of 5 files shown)


sashei:
for code reviewer:

lots of refactoring in `ItemListPresenter`, effective changes are at lines `283` and `329`.

devinreams:
This is looking _real_ good now. Thanks @sashei.

I tried both the "empty appletester" account (needing confirmation) and the "full appletest" account (no confirmation needed) to see the combination of states and I think this gets us what we wanted.

@mozilla-lockbox/ux can you also please take a look and test this out?

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -371,3 +332,75 @@ extension ItemListPresenter {
         }
     }
 }
+
+extension ItemListPresenter {
```
linuxwolf: just curious: why have an extension for `ItemListPresenter` instead of adding these functions to the class directly?

sashei: I've been separating the `fileprivate` smaller functions into their own extension for organizational purposes :p

sashei:
@isabelrios looks like this is failing because of the test account -- did you ever get an update on that? I'd like to use a similar one for the automated screenshots.

PR #481, opened 2018-06-13
Long press on detail screen

### LINKED ISSUE #477: Long press should also copy username/password
In addition to "tap" to copy, can we just add "long tap" or "press and hold" or whatever other combinations of tapping happen?

We've observed users tap and hold, much like iOS has you do when selecting text, and not getting anything/feedback.

joeyg:
Resolves #477. I didn't come up with a great way to add xctest's for this - maybe better done with XCUITest?

### DIFF
--- a/lockbox-ios/View/ItemDetailView.swift
+++ b/lockbox-ios/View/ItemDetailView.swift
@@ -51,6 +51,7 @@ class ItemDetailView: UIViewController {
     @IBOutlet weak var tableView: UITableView!
     @IBOutlet weak var learnHowToEditButton: UIButton!
     var itemId: String = ""
+    let longPress = UILongPressGestureRecognizer()
 
     override var preferredStatusBarStyle: UIStatusBarStyle {
         return UIStatusBarStyle.lightContent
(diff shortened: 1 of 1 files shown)


changecourse:
@devinreams yes I think we'd want functional parity for both the username and password fields, regarding long press functionality.

joeyg:
oops - this was supposed to work for both fields. If it does not then i'll try and redo it.

devinreams:
>this was supposed to work for both fields. If it does not then i'll try and redo it.

Yeah it's odd. I'm staring at the view and presenter code and not sure why it wouldn't Just Work® on the copyable username field too. I've exhausted my ability to help troubleshoot, sorry. 😔

joeyg:
@devinreams updated the PR to set the gesture recognizer on the tableview instead of the cells. Let me know if this works better.

devinreams:
>updated the PR to set the gesture recognizer on the tableview instead of the cells

That did the trick 👍 

I'm doing a bit more hammering 🔨 to make sure there aren't unintended side-effects. Otherwise this is good to go. Thanks again @joeyg 💝

PR #483, opened 2018-06-14
scroll to top on sort - WIP

### LINKED ISSUE #439: Reset scroll position when the sort order changes
Really minor issue.
STR:

1. Scroll down
2. Change the sort order from A-Z to "Recent"

Now I'm basically at a random point in time (not withstanding #416) so it would be good to reset the scroll position whenever the sort order changes.

joeyg:
Resolves #439.

### DIFF
--- a/lockbox-ios/Presenter/ItemListPresenter.swift
+++ b/lockbox-ios/Presenter/ItemListPresenter.swift
@@ -17,6 +17,7 @@ protocol ItemListViewProtocol: class, AlertControllerView, SpinnerAlertView {
     var tableViewScrollEnabled: AnyObserver<Bool> { get }
     func dismissKeyboard()
     var pullToRefreshActive: AnyObserver<Bool>? { get }
+    func scrollToTop()
 }
 
 struct LoginListTextSort {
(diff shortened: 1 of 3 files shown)


lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -116,12 +117,20 @@ class ItemListPresenter {
     lazy private var alphabeticSortObserver: AnyObserver<Void> = {
         return Binder(self) { target, _ in
             target.settingActionHandler.invoke(SettingAction.itemListSort(sort: ItemListSortSetting.alphabetically))
+
+            if let view = self.view {
+                self.view?.scrollToTop()
```
joeyg: This might belong inside a subscribe to the setting observer. not sure.

devinreams:
I just tested on iPhone X and saw some interesting behavior:

1. the first re-sort scrolled to the top, but with the first item hidden, so not sure if the offset is calculated from the bottom of the first item or has to do with the filter bar or something?

2. subsequent re-sorts seemed to scroll to top (immediately) but then back down to the existing position (animated)

Steps to reproduce: open Lockbox, scroll halfway down, change sort order, change sort order again

PR #491, opened 2018-06-21
dynamic text support

### LINKED ISSUE #470: most buttons and text labels do not respect and adjust to "larger text" / dynamic sizes
From #441 

### Text Style Definitions
https://mozilla.invisionapp.com/share/3ALGEQ8RDM9#/304386170_Text-Styles

Combined Logo & Tagline Image:
[logo.zip](https://github.com/mozilla-lockbox/lockbox-ios/files/2121236/logo.zip)

sashei:
resolves #470


devinreams:
@mozilla-lockbox/product I will test this myself but can one of you take a look, too?

Cool if not, just hoping to merge early this week for final testing before re-submission.

devinreams:
This is great! Thanks for this @sashei.

I spotted just a couple little things that looked unexpected or maybe unintentional when using the default font size (so to me they felt like regressions, but acknowledge may be trade-offs based on the new sizes we're using, I defer to you and the UX team):

1. The "Search" magnifying glass placeholder fell off to the left and isn't padding to the right like it used to be
  iphone_8_-_11_4

2. The currently-set Settings values are quite a bit smaller than the original size
  iphone_8_-_11_4

And a few other things from experimenting with other sizes, likely lower priority and to look at after launch, but wanted to capture:

3. The "Learn More" button at the bottom of entry detail has an arrow which is spaced with a hard-coded size/offset and sits over the text
  iphone_8_-_11_4

4. I think the layout or dynamic size of the "Username copied" box will get too cramped. 
  banners_and_alerts_and_iphone_8_-_11_4

5. The actual text in the detail view's address/username/password cells does not appear to grow in size (the cell labels do)
  screen shot 2018-06-25 at 3 03 48 pm

6. The "Lock Now" button does not seem to adjust to all sizes
  screen shot 2018-06-25 at 3 06 20 pm

I was amazed and glad to see the app is still completely usable (janky but usable) at the very smallest and very largest sizes. Nice work! 👍

lockbox-ios/View/ItemListView.swift
```diff
@@ -321,6 +325,9 @@ extension ItemListView {
         let searchImageView = UIImageView(frame: CGRect(x: 0, y: 0, width: 40.0, height: cell.frame.height))
         searchImageView.contentMode = .center
         searchImageView.image = searchImage
+        if #available(iOS 11.0, *) {
+            searchImageView.adjustsImageSizeForAccessibilityContentSizeCategory = true
```
devinreams: I think this pushed the image over to the left and looks cut off on the edge (should have some padding and shift back right?)

lockbox-ios/View/SettingListView.swift
```diff
@@ -73,6 +73,9 @@ extension SettingListView {
                 cell.accessibilityCustomActions = cellConfiguration.accessibilityActions
                 cell.accessibilityIdentifier = cellConfiguration.text
 
+                cell.textLabel?.font = UIFont.preferredFont(forTextStyle: .body)
+                cell.detailTextLabel?.font = UIFont.preferredFont(forTextStyle: .caption2)
```
devinreams: Caption2 feels really small. Its correct for the "Send Usage Data" label, but I think unexpected for the "Auto Lock" and "Preferred Browser" settings.

sashei:
@nickbrandt can I get the asset for the `ItemDetail` learn more button (a right-pointing arrow) in white instead of black?

sashei:
Also, I've discovered another place where the font is different... the mono-space font we use for the password field. What do we think we should do in this case @changecourse ?

changecourse:
@sashei here's the arrow. 

[arrow.zip](https://github.com/mozilla-lockbox/lockbox-ios/files/2137981/arrow.zip)

As for the mono-space font... hmmmm, ideally we'd want to keep that preserved and just match sizing changes, small to large, for the 17px username field above. can we have that be a 'one-off' custom class?

sashei:
ok this is ready for re-review

sashei:
note: everything except for the monospace font unfortunately, I am going to open another issue #506 for that and move on the the autolock bug for now

devinreams:
@sashei Thanks for cleaning all of those things up! 🛀 

Last thing: I noticed the second bullet point on the onboarding confirmation screen (encryption) was not dynamically sized. It appears to be a text view, not a label like the other two. Should that be swapped out?

Everything else looks good to ship!

changecourse:
Approving, with caveats below... 

screen shot 2018-06-27 at 9 37 58 am
Confused about why icons are behaving in that manner, as well as truncation acting two different ways

screen shot 2018-06-27 at 9 38 23 am
"Lock Now" button not getting bigger relative to font size, also losing padding from bottom

screen shot 2018-06-27 at 9 38 36 am
Why is search truncating so early, relative to the space available?

screen shot 2018-06-27 at 9 38 48 am
"Learn how to edit this entry" appears to be fixed to the base of the screen, overlapping with the rest of the content on the detail...

devinreams:
Thanks @changecourse.

Since we all agreed a few days ago that this first pass is to add the dynamic size support (better to have something than nothing) and make sure obvious stuff is sorted out, I filed a follow-up issue #512 to fix these kinds of situations (as there may be _many_, like you readily found at the hugest size).

I finished up testing with the "reasonable" increased sizes and don't see anything else that makes the app unusable. I say we merge now so we can get this in the hands of internal testers ASAP.

devinreams:
Thanks for all your hard work on this one, @sashei.

As we've all discussed before but worth repeating: supporting these dynamic sizes is one of those things that will _really_ help those that need it the most. Thanks again. ❤️

PR #508, opened 2018-06-26
flip constraints so "Learn more" anchors to bottom, not "Get Started" button - WIP

devinreams:
this way as we get taller screens, its easier for auto layout to keep the "learn more" and blurb down towards the bottom of the screen, not awkward floating around the belly button...

this also matches how we anchor other "learn more" links like on the item detail view.

## Before

Fine on iPhone SE, but kinda hanging out in the middle of the screen on iPhone X and taller screens

screen shot 2018-06-26 at 3 50 52 pm

## After

screen shot 2018-06-26 at 3 48 39 pm
screen shot 2018-06-26 at 3 49 21 pm


devinreams:
Closing as stale and it was a last minute idea I had before launch.

Will wait for input from UX if we want to change any layouts in future updated...

PR #513, opened 2018-06-27
Ignore XCUITests from code coverage

### LINKED ISSUE #510: Code coverage reduced to 88%
Looks like the big culprit is the XCUITests. Should those have more coverage? Or be excluded from codecov? Or something else?

devinreams:
Fixes #510

Is this cool @sashei @isabelrios? Or should we have test coverage on the tests themselves?

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,4 @@
 ignore:
-    - "lockbox-iosTests/*"
+    - "LockboxXCUITests/*"
     - "lockbox-ios/Store/DataStore.swift"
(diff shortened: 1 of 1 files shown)


devinreams:
Oh it may be that codecov is getting the _last_ report from the tests being run, so its report not the unit tests but the UI tests since they happen last: https://codecov.io/gh/mozilla-lockbox/lockbox-ios/tree/master

The tree only shows the files for the XCUITests from what I can tell. So maybe more codecov configuration needed here. 🤔

PR #527, opened 2018-06-29
route to list if booting while waiting for email confirmation

### LINKED ISSUE #518: User forced to login a second time if hard quitting before confirming email
Steps to reproduce:

1. Login to an FxA account that you need to confirm (you can either tap `FINISH` on the welcome onboarding page or not, this bug will occur either way)
2. Hard quit the app, don't confirm the email
3. Re-launch the app

expected: still in `Welcome to Firefox Lockbox` onboarding page or entries list with email confirmation message
actual: back to `Get Started` login page

### LINKED ISSUE #552: "Welcome to Lockbox" confirmation screen auto-dismissed without tapping "Finish"
One our our participants in a UR session went through Lockbox for the first time on her device, and successfully authenticated; She was briefly presented with the confirmation screen (approx 2-3 seconds) but it auto-dismissed without her tapping on the "Finish" button.

That being said, I haven't been able to reproduce it on my current physical/virtual device(s).

sashei:
resolves #518 
resolves #552

### DIFF
--- a/lockbox-ios/Action/RouteAction.swift
+++ b/lockbox-ios/Action/RouteAction.swift
@@ -15,6 +15,10 @@ class RouteActionHandler: ActionHandler {
     func invoke(_ action: RouteAction) {
         self.dispatcher.dispatch(action: action)
     }
+
+    func invoke(_ action: OnboardingStatusAction) {
+        self.dispatcher.dispatch(action: action)
+    }
 }
 
 protocol RouteAction: Action { }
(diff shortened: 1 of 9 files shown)


devinreams:
I think we should get one more user experience review here.

Especially given the descriptions at #518, I want to make sure we get a 👍 before merging.

sashei:
@linuxwolf requesting re-review because of new code; I think we will need the `OnboardingStatus` when rebasing the OAuth branch with these changes because we lose the `preparing` state.

PR #529, opened 2018-06-30
App version

### LINKED ISSUE #528: Output version and build number within the app
Idea from @teonbrooks: for end users to easily troubleshoot and know which version they're running.

I've seen other apps add it to a settings or similar screen. See Slack for instance in the Settings table cells:

img_3283

joeyg:
Resolves #528.

simulator screen shot - iphone 8 - 2018-06-30 at 13 54 26


devinreams:
@mozilla-lockbox/ux what do you think about:

1. dropping the version in to a super obvious table row like this?
2. the label (App Version)
3. the location (in the Support section)
4. anything else?

👍 👎

PR #537, opened 2018-07-11
Pause autolock while the user is on a webpage.

### LINKED ISSUE #534: App locks when interacting with webviews
A redditor pointed out that they had issues with the app locking while providing feedback as seen [here](https://www.reddit.com/r/firefox/comments/8xox1z/firefox_lockbox_for_ios_and_notes_for_android_are/e24peq5/).

I just confirmed this happens. It seems the interactions in the webview don't keep the autolock timer active:
- set the autolock timer to 1 minute
- navigate to the feedback form
- actively engage with the form for 1 minute
- the app locks while interacting with the form

I see this being a potentially serious issue for us. Even with the default timer of 5 minutes, we're going to see feedback submissions drop which will likely impact our app store rating in a negative way. If the timer is set to 1 minute, it's unlikely someone will complete the feedback form in that time.

jhugman:
Fixes #534.

This PR adds a `paused` property to the `AutoLockStore`.

The auto-lock timer is paused when the user goes to webpage. 

If the timer has already expired when the user closes the page, then the app is locked.

Otherwise, the timer is reset as is normal. 

This allows the user to spend as long as they want on the FAQ or feedback page, without playing whackamole catching events out of a webpage, or compromising the security of the app — sample attack: pausing the timer by going to the FAQ page, then quickly reseting the autolock time to `Never`.

### DIFF
--- a/lockbox-ios/Store/AutoLockStore.swift
+++ b/lockbox-ios/Store/AutoLockStore.swift
@@ -16,6 +16,7 @@ class AutoLockStore {
     private let userDefaults: UserDefaults
 
     var timer: Timer?
+    var paused: Bool = false
 
     init(dispatcher: Dispatcher = Dispatcher.shared,
          dataStore: DataStore = DataStore.shared,
@@ -52,12 +53,19 @@ class AutoLockStore {
                     self?.resetTimer()
                 })
                 .disposed(by: self.disposeBag)
+
+        self.dispatcher.register
+            .filterByType(class: ExternalWebsiteRouteAction.self)
+            .subscribe(onNext: { [weak self] _ in
+                self?.pauseTimer()
+            })
+            .disposed(by: self.disposeBag)
     }
 }
 
 extension AutoLockStore {
     private func resetTimer() {
-        self.stopTimer()
+        self.stopTimer(reset: !paused)
         self.setupTimer()
     }
 
@@ -86,7 +94,7 @@ extension AutoLockStore {
 
     private func setTimer(seconds: Int) {
         let timerValue = self.userDefaults.double(forKey: SettingKey.autoLockTimerDate.rawValue)
-        if timerValue != 0 && timerValue > Date().timeIntervalSince1970 {
+        if timerValue != 0 && timerValue < Date().timeIntervalSince1970 {
             self.timer = Timer(fireAt: Date(timeIntervalSince1970: timerValue),
                     interval: 0,
                     target: self,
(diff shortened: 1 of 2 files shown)


lockbox-ios/Store/AutoLockStore.swift
```diff
@@ -86,7 +94,7 @@ extension AutoLockStore {
 
     private func setTimer(seconds: Int) {
         let timerValue = self.userDefaults.double(forKey: SettingKey.autoLockTimerDate.rawValue)
-        if timerValue != 0 && timerValue > Date().timeIntervalSince1970 {
+        if timerValue != 0 && timerValue < Date().timeIntervalSince1970 {
```
linuxwolf: this makes me wonder how this was ever working ...

joeyg: It seems like this should trip up the unit tests. It seems like this will create an invalid timer.

jhugman: It worked because we were always calling `stopTimer()` right before `setTimer()`, so `timerValue` was _always_ 0.

With this new change, the timer is invalidated, but the `timerValue` in `userDefaults` is not zeroed.

Top tip: never use `>`.

devinreams:
@jhugman Heads up that we are struggling to test and confirm additional timers (15 minutes and 30 minutes) work as expected at #551 and suspect there may be a regression introduced here where autolocking is just not firing.

For example, simply installing the app and leaving it open for 5 minutes (default timer) does not then lock after 5+ minutes...

I'll open a new issue if that's the case and consider troubleshooting fixing the issue...

devinreams:
I did some `git bisect` to try and isolate if something broke with the timer and can't concretely point any fingers right now. 👇 👈 

I'm not sure if I can consistently reproduce an issue with a _brand new_ app installation, which is good. I delete the app and fully re-install (on device), I reset the iOS Simulator and the default 5 minute timer works ✅  and if I change to 1 minute timer it works. ✅ 

And if I install an old pre-#537 build (Build 1574) and then upgrade to a post-#537 build (Build 1644) then it still works.

- Install Build 1574
- sign in to FxA
- wait 5 minutes
- app locks ✅ 
- Install Build 1644 over top of it
- wait 5 minutes
- app locks ✅ 
- change timer setting to 1 minute
- app locks ✅  

Sorry for the noise... just wanted to capture all the troubleshooting related to #551 I was doing today.

PR #547, opened 2018-07-12
Add support for Klar browser

### LINKED ISSUE #540: Include Klar as a browser option for DE
In our German market, Firefox Focus is not a browser option. Instead the privacy browser we can provide is Klar.

When we're looking to offer the app in Germany we'll want to replace the Firefox Focus browser option with Klar.

joeyg:
Resolves #540.

### DIFF
--- a/lockbox-ios/Presenter/PreferredBrowserSettingPresenter.swift
+++ b/lockbox-ios/Presenter/PreferredBrowserSettingPresenter.swift
@@ -21,8 +21,6 @@ class PreferredBrowserSettingPresenter {
     lazy var initialSettings = [
         CheckmarkSettingCellConfiguration(text: Constant.string.settingsBrowserFirefox,
                                           valueWhenChecked: PreferredBrowserSetting.Firefox),
-        CheckmarkSettingCellConfiguration(text: Constant.string.settingsBrowserFocus,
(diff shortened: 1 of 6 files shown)


lockbox-ios/Presenter/PreferredBrowserSettingPresenter.swift
```diff
@@ -50,6 +48,10 @@ class PreferredBrowserSettingPresenter {
     }
 
     func onViewReady() {
+        if let browser = getInstalledFocusBrowser() {
```
devinreams: Nice 👍

devinreams:
Code looks good and still works as expected for me. 👍 

- ✅ I tested locally with the Simulator and it shows Focus (since neither it nor Klar can be installed).
- ✅ I also tested on iPhone X and it shows Focus (and it is installed).
- ⚠️ I did not test with Klar since I don't have a German store account, but we can ask some German folks (after this is merged) to test and confirm via TestFlight, but the [code here matches the expected URL scheme](https://github.com/mozilla-mobile/focus-ios/blob/518cfb06b0c03b562f027f8e3dd8df05f5e2635a/OpenInFocus/ActionViewController.swift#L38) so.. it _should_ just work.

Thanks @joeyg. 💖

PR #549, opened 2018-07-13
Lock Now button for users without passcode

### LINKED ISSUE #548: Bring back "Lock Now" for users that do not have biometrics or passcode set.
In order to keep users informed on how they can secure access to their logins with locking, we have decided to bring back the "Lock Now" option even for those who **do not** have biometrics or passcode setup.

When a user without biometrics or passcode setup on their device taps "Lock Now", we would like to display a dialog informing them to set up a device passcode in order to lock the app.

_Usability session notes of participant without device auth set up_
https://docs.google.com/document/d/1KAuLZGvag0Xs4A6YB_eXEAaHLySEBu4hCQaK3FoPK2g/edit?usp=sharing

### Visual Design
**InVision**
https://mozilla.invisionapp.com/share/BQEUSO0S2#/308670491_Settings_-_Lock_-_No_Passcode

**Zeplin**
https://zpl.io/V0yKDko

joeyg:
Resolves #548.


devinreams:
@nickbrandt how do you tend to test this?  With a real device right?

I don't have a good test device and can't "simulate" no passcode on the Simulator so will have to ask for you to test this (and/or your test device if you want me to check it out, too).

nickbrandt:
@devinreams Yep, I use a real device. I tested it on my phone and it works as intended. @changecourse has the test device if either you/he wants to test as well.

joeyg:
Thanks for the help testing @changecourse @nickbrandt !

PR #583, opened 2018-07-30
Disable scrolling only when syncing an empty login list

### LINKED ISSUE #445: "Pull to refresh" doesn't work when the logins list is empty
v1.0 (1343)
1. Create a new account in Lockbox iOS.
2. Connect the account on desktop.
3. Save a login on desktop.
4. Try to manually trigger (pull to refresh) a sync in Lockbox to have the previously created login transferred.

Expected: Pull to refresh should work when the logins list doesn't have any entries.

Actual: Pull to refresh doesn't work when the logins list is empty.

sigito:
Fixed #445.


devinreams:
Thank you again for your contribution, @sigito 👏 👏

PR #618, opened 2018-08-06
update xcode version

joeyg:
Add XCode version to install documentation. Installing with a prior version of swift requires building dependencies from source which requires rust.


devinreams:
I'm sorry you got snagged in an un-workable spot but glad the resolution was fairly easy. Thanks for wading through this and figuring it out, @joeyg. ❤️

PR #633, opened 2018-08-23
make sure to preserve current version even when clearing keychain

### LINKED ISSUE #631: Reauthentication Required on every launch
Every time I build and launch a simulator I get `Reauthentication Required` dialog.

sashei:
resolves #631 

note: to replicate this bug, you simply need to install the upgrade story (I believe over a build that already had OAuth, but possibly older ones), see the upgrade dialog, and authenticate.

### DIFF
--- a/lockbox-ios/Common/AppDelegate.swift
+++ b/lockbox-ios/Common/AppDelegate.swift
@@ -16,8 +16,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
 
-    let keychainWrapper = KeychainWrapper.standard
-
     func application(_ application: UIApplication, willFinishLaunchingWithOptions
                      launchOptions: [UIApplicationLaunchOptionsKey: Any]? = nil) -> Bool {
         if !isRunningTest {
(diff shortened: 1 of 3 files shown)


nickbrandt:
@sashei This is resolving the issue where it required me to log in every time I force quit, however, I am now no longer getting the prompt when updating. It simply directs me to the Welcome screen.

Disconnected from version w/ OAuth > Deleted App > Installed build 1717 (pre OAuth) > Updated to 1951 > Directed to Welcome Screen (no prompt)

devinreams:
I was seeing the same issue described at #631 when I went from 1.1.1 (via App Store) to 1.2 (via TestFlight).

So I re-tested by:

- installing Build 1717 (via production branch on buddybuild)
- signing in and locking the app
- installing Build 1961 (via this branch on buddybuild)
- launching the app and getting the prompt ✅ 
- signing in 💥 (my app crashed too now)
- re-opening the app and was signed in looking at logins in list ✅ 

So I re-tried again by disconnecting and deleting the app, then:

- install 1717
- sign in
- install 1961
- get the prompt ✅
- tap to continue 
- sign in (no crash)
- looking at logins in list ✅ 

@sashei Want to merge this then push to TestFlight (to just us) and try to catch the crashes?

Issue #486, opened 2018-06-20
Autofill Extension (iOS 12)


devinreams:
Follow up to #476 exploration

This is an epic where we want to capture the consolidated design/user stuff for the autofill feature, we may split out to separate engineering stories as needed.

- Project / Autofill Credential Provider configuration
-`ASCredentialIdentityStore` to store and update data
- List view
  - no need for detail entry view
  - high potential for reuse, minimal stripped-away design
- Minimal authentication view (to unlock with Face ID / Touch ID / passcode)
- User onboarding to explain autofill during first run of Lockbox app
- Telemetry events
- More content in FAQ
- Link from our settings screen to get to autofill settings

## Designs

**InVision**
Onboarding Flow to Settings
https://mozilla.invisionapp.com/share/BQEUSO0S2#/316626649_AutoFill_-_Onboarding

Example App Flow for AutoFill
https://mozilla.invisionapp.com/share/BQEUSO0S2#/316626653_AutoFill_-_App_Example

## Epic

- parent of #609
- parent of #610
- parent of #611
- parent of #612
- parent of #613
- parent of #614
- parent of #615
- parent of #616
- parent of #617
- parent of #624
- parent of #626

## Acceptance Criteria

- [x] Should be prompted during onboarding to set up autofill for apps and websites (direct user to the specific iOS autofill setting to turn on “Firefox Lockbox”)
- [x] Should be able to get back to “Firefox Lockbox” app after turning on autofill (if not able, then instructions to be provided in onboarding) 
- [x] Should have the option to set up autofill from settings
- [x] Should present user with AutoFill suggestions in the quicktype bar when focusing on a username or password from within safari or an app.
- [ ] Should highlight the domain associated in the Credential List with the ability to view all credentials.
- [x] Should display a Credential List that prioritizes filling (the only action to take from this view)
- [x] If no match is found in the Credential List, should display all options with some ability to search/filter
- [x] Should not be able to go to the credential details
- [x] Should not be able to go to settings from the Credential List
- [x] Should have a clear way to cancel (choose not to select a login)
- [x] When selecting a quicktype bar suggestion, should present the user with a choice to authenticate to autofill with those credentials (if Lockbox is locked)
- [x] When selecting the passwords link (key icon), should present the user with a choice to authenticate in order to view the credential list.
- [ ] Should have the ability to set specific Lockbox autofill settings (ex: quicktype bar suggestions) - this is a big maybe and requires some exploration

[...]

PR #639, opened 2018-08-24
Autofill!

sashei:
Connected to #486

With the addition of the Credential Provider API in iOS 12, you will need to use **Xcode 10** (beta 6) and **Swift 4.2** (`swiftlang-1000.0.32.1 clang-1000.10.39`) to work on or build this branch.

This will serve as the long-running feature branch until we can work on `master` directly due to Xcode upgrades.

Remaining todos:

- [x] test and merge #666


devinreams:
Oh yes! Please tell me you'd like to do Pull Requests _into this branch_ for all the other little feature branches 🤞 I love branching strategy...

sashei:
as much as it pains me to say it.... yes.

sashei:
@irrationalagent the `CredentialProviderPresenter.swift` is the rough equivalent of the `RootPresenter` in Lockbox.app. There is a similar `Action` cycle in this application though, I might hold tight for the next PR to come in (#649) because that will have more infrastructure for Telemetry-interesting actions in place.

CredentialProvider/View/CredentialProviderView.swift
```diff
@@ -0,0 +1,86 @@
...
+     Prepare your UI to list available credentials for the user to choose from. The items in
+     'serviceIdentifiers' describe the service the user is logging in to, so your extension can
+     prioritize the most relevant credentials in the list.
+    */
+    override func prepareCredentialList(for serviceIdentifiers: [ASCredentialServiceIdentifier]) {
+        
+    }
+
+    override func provideCredentialWithoutUserInteraction(for credentialIdentity: ASPasswordCredentialIdentity) {
+        let databaseIsUnlocked = true
```
jhugman:
```swift
// Dummy implementation
```

Shared/Common/Extensions/ASCredentialIdentityStore+.swift
```diff
@@ -0,0 +1,16 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import AuthenticationServices
+
+@available(iOS 12, *)
+protocol CredentialIdentityStoreProtocol {
+    func getState(_ completion: @escaping (ASCredentialIdentityStoreState) -> Void)
+    func removeAllCredentialIdentities(_ completion: ((Bool, Error?) -> Void)?)
```
jhugman: Suggest using a typealias here.
e.g. 

```swift
typealias CredentialCompletion = ((Bool, Error?) -> Void)
```

Then use it as optional in the method parameter.

```swift
func removeAllCredentialIdentities(_ completion: CredentialCompletion?)
```

Shared/Store/BaseDataStore.swift
```diff
@@ -0,0 +1,438 @@
...
+        case (_, _, NotificationNames.ProfileDidStartSyncing):
+            // fall through for the locked and unlocked states.
+            syncSubject.onNext(.Syncing)
+        case (_, _, NotificationNames.ProfileDidFinishSyncing):
+            // end of all syncs
+            syncSubject.onNext(.Synced)
+        default:
+            print("Unexpected state combination: \(storageState) | \(syncState), \(notification.name)")
+        }
+    }
```
jhugman: Cherrypicking is going to be annoying here.

Shared/Store/BaseDataStore.swift
```diff
@@ -0,0 +1,438 @@
...
+        }
+
+        func resetProfile() {
+            self.profile = profileFactory(true)
+            self.initializeProfile()
+            self.syncSubject.onNext(.NotSyncable)
+            self.storageStateSubject.onNext(.Unprepared)
+        }
+
+        stopSyncing() >>== disconnect >>== deleteAll >>== resetProfile
```
jhugman: Cherrypicking is going to be annoying here.

Shared/Store/BaseDataStore.swift
```diff
@@ -0,0 +1,438 @@
...
+                .subscribe(onNext: { autoLockSetting in
+                    switch autoLockSetting {
+                    case .Never:
+                        self.unlock()
+                    default:
+                        let date = NSDate(
+                                timeIntervalSince1970: self.userDefaults.double(
+                                        forKey: UserDefaultKey.autoLockTimerDate.rawValue))
+
+                        if date.timeIntervalSince1970 > 0 && date.timeIntervalSinceNow > 0 {
```
jhugman: Prefer to remove `>`.

oh, wait this is to make sure we have an `autoLockTimerDate`.

```swift
if let seconds = self.userDefaults.double(forKey: UserDefaultKey.autoLockTimerDate.rawValue),
     seconds > 0 {

} else {

}
```

Actually, I'm not sure what this code is supposed to do, or that it wasn't looked at in another PR.

lockbox-ios/Store/AccountStore.swift
```diff
@@ -8,57 +8,65 @@ import RxCocoa
...
-    static let allValues: [KeychainKey] = [.accountJSON, .email, .displayName, .avatarURL]
+import Shared
+
+/* These UserDefault keys are maintained separately from the `UserDefaultKey`
+ * enum in BaseConstants.swift because, for the time being, they are only
+ * useful in the main application. However, rather than maintaining two
+ * separate UserDefaults instances, all the values for these keys are stored
+ * in the app group instance so that no further migrations will be required
+ * in the case that they become relevant in an app extension context. */
+enum LocalUserDefaultKey: String {
```
jhugman: I'm not sure I understand the need for two sets of keys, especially if there is scope for sharing the keys.

devinreams:
Quick status update:

- three uispecs are failing ❌ but known and understood 👌 
  - two of them (3 and 6) are kinda flakey and being addressed at #686 
  - one is because we introduced a screen before the welcome (0)

- @jhugman has requested some nits/changes

- shortly after this message, this branch will require **beta 6 of Xcode 10** (up from beta 5)

CredentialProvider/Presenter/CredentialWelcomePresenter.swift
```diff
@@ -0,0 +1,94 @@
...
+
+    private func populateCredentials() {
+        self.dispatcher.dispatch(action: CredentialProviderAction.refresh)
+
+        let populated = self.credentialProviderStore.state
+            .filter { $0 == CredentialProviderStoreState.Populated }
+            .map { _ -> Void in return () }
+
+        self.credentialProviderStore.state
+                .asDriver(onErrorJustReturn: CredentialProviderStoreState.NotAllowed)
```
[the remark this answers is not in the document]
sashei: It's used later for onboarding ... I almost did though, you're right.

Shared/Common/Extensions/ASCredentialIdentityStore+.swift
sashei: doing this messes up the injection pattern that these are useful for :(

Shared/Store/BaseDataStore.swift
sashei: this is from another PR.

lockbox-ios/Store/AccountStore.swift
sashei: It has to do with minimizing what settings are included in the extension. eg, we don't need to share the preferred browser setting in the extension.

PR #641, opened 2018-08-24
[WIP] point to new fxa-sync code, disconnect fully

### LINKED ISSUE #640: Start up crash due to FxAUtils
browser profile related. see Build 1940 crash logs...

sashei:
resolves #640


sashei:
@devinreams I am curious what your experience disconnecting first from the OAuth branch would be.

I am also still seeing "No matching entries" on disconnect / reconnect so I think that bug is still present.

sashei:
just updated from master because I saw a bug that we have already fixed.

sashei:
...wait getting same results as @devinreams now...

devinreams:
OK, after a quick chat we're gonna pause on this and pick it back up later. Moving on to autofill stories and will submit the 1.2 release as-is.

sashei:
This approach doesn't work; to revisit!

PR #643, opened 2018-08-24
Add a "Get Support" link to Discourse in setting list

### LINKED ISSUE #637: Provide support in the app
Right now we don't provide users a way to ask a question or report an issue. While many of the 'feedback' submissions do include these, it's a less than ideal process.

## Acceptance criteria
* Should have a clear link for "support" within the app itself that is distinctly different than "feedback"
* Should link to our [Discourse forum](https://discourse.mozilla.org/c/test-pilot/lockbox) so that anyone on our team can respond
* Should prompt for details: device, OS, and screenshots where possible

devinreams:
Resolves #637 (still marked as "needs-design")

I took some liberty here and assumed we'd just add a link in the existing Support section and called that link "Ask Questions" (as a placeholder).

To test:

- open app and sign in
- open Settings screen
- observe the link in the "Support" heading
- tap it and it opens a web view with the discourse forum

Screenshots demonstrating this:

screen shot 2018-08-24 at 4 49 37 pm

screen shot 2018-08-24 at 4 49 53 pm

### DIFF
--- a/lockbox-ios/Action/TelemetryAction.swift
+++ b/lockbox-ios/Action/TelemetryAction.swift
@@ -40,6 +40,7 @@ enum TelemetryEventObject: String {
     case settingsItemListSort = "settings_item_list_sort"
     case settingsFaq = "settings_faq"
     case settingsProvideFeedback = "settings_provide_feedback"
+    case settingsGetSupport = "settings_get_support"
     case loginWelcome = "login_welcome"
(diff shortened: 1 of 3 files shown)


changecourse:
I think we should change it from plural to singular, so the call-out reads "Ask a Question" 

Otherwise, placement and approach LGTM!

PR #644, opened 2018-08-24
Replace references to missing arrowhead

### LINKED ISSUE #627: [ItemList] "arrowhead-right-16" image is not part of the project
Referenced in "Learn more" button image `arrowhead-right-16` is not included in the repo.

> Could not load the "arrowhead-right-16" image referenced from a nib in the bundle with identifier "org.mozilla.ios.Lockbox"

devinreams:
Fixes #627 

The image asset was moved so the storyboard needed to be updated to use those arrows.

- item list: no entries
- item list: no matching entries

I also shrunk this from 22px tall down to 16px. It still seems a bit tall but that may just be me.

This is also used in other places like on the item detail view at the bottom so I checked to see it still looks ok there.

Here's what it looks like:

screen shot 2018-08-24 at 5 00 32 pm
screen shot 2018-08-24 at 5 01 02 pm


changecourse:
While we have the arrow within a 12x12 bounding box in our comps, in actuality, it is sized as 8px high for 1x density environments, in order to align with the corresponding font size of the callout. 

screen shot 2018-08-27 at 10 02 57 am

devinreams:
>While we have the arrow within a 12x12 bounding box in our comps, in actuality, it is sized as 8px high for 1x density environments

@changecourse OK, that helps, thanks.

I've moved things around on the list view and item detail, then resized the images a bit more, so these should look (closer to?) as expected.

screen shot 2018-08-27 at 10 23 23 am
screen shot 2018-08-27 at 10 25 56 am

Issue #659, opened 2018-08-28
Track how often users "Disconnect"


devinreams:
Likely removed as there are reset events in Telemetry but we don't seem to dispatch the reset anymore.

Historical context in Slack (private): https://mozilla.slack.com/archives/C4LNYNJ8G/p1535489611000100

PR #660, opened 2018-08-29
add login and disconnect oauth telemetry events

irrationalagent:
resolves https://github.com/mozilla-lockbox/lockbox-ios/issues/659

This seems to work for me locally, but I wasn't sure how to handle the `oauthSignInMessageRead` case - we don't really need an event here but I don't know how to elegantly handle that in swift. If its better to just include it, then that's OK. 

BTW I didn't experience any bugs during or after having to re-login with oauth in the simulator, fwiw. (I hadn't done it since the code landed in master)

Sasha I know you're at a conference, no need to do this until you're back

### DIFF
--- a/lockbox-ios/Action/AccountAction.swift
+++ b/lockbox-ios/Action/AccountAction.swift
@@ -29,3 +29,28 @@ extension AccountAction: Equatable {
         }
     }
 }
+
+extension AccountAction: TelemetryAction {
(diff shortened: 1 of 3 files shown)


irrationalagent:
just realized i need to write a test for this, ill do it tomorrow

devinreams:
Thanks @irrationalagent, we'll get this into the point release so we can have (potentially) actionable data. 🤘

PR #664, opened 2018-08-29
preserve settings and authentication on upgrade to autofill-enabled build

### LINKED ISSUE #626: Preserve user's settings and authentication state when moving to Autofill
Currently, we just assume that everything "exists" in the shared UserDefaults and Keychain world with the Autofill extension. 

However, all users on upgrade will still have their settings and authentication stored in the app-only versions. When we detect the upgrade, we need to read from the app-only Keychain and UserDefaults into the shared ones.

depends on #496 
depends on #609

sashei:
resolves #626 

## test steps

1. install a recent version of the app from buddybuild 
  - probably best to pick the latest from the production branch (1.2 = 2051)
  - "upgrading" in this case refers to upgrading from a 1.2 release build. `master` builds should work just fine in this case too
  - note: for a complete end-to-end test, `Disconnect` from an Autofill build if you have previously installed one before downgrading.

2. sign in using FxA

3. change settings (browser, autolock time) in the app and remember what they are

3. install a build from this branch

**testing notes**: 

- you should see nothing unexpected when upgrading and launching lockbox (i.e., aren't forced to sign in again).
- settings (such as autolock time & send usage data) are preserved when upgrading.

### DIFF
--- a/lockbox-ios/Store/UserDefaultStore.swift
+++ b/lockbox-ios/Store/UserDefaultStore.swift
@@ -39,6 +39,19 @@ class UserDefaultStore: BaseUserDefaultStore {
                 })
                 .disposed(by: self.disposeBag)
 
+        self.dispatcher.register
+                .filterByType(class: LifecycleAction.self)
+                .subscribe(onNext: { action in
+                    guard case let .upgrade(previous, _) = action else {
+                        return
+                    }
+
+                    if previous <= 2 {
+                        self.readValuesFrom(UserDefaults.standard)
+                    }
+                })
+                .disposed(by: self.disposeBag)
+        
         self.loadInitialValues()
     }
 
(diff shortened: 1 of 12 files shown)


lockbox-ios/Store/UserDefaultStore.swift
```diff
@@ -60,3 +73,19 @@ class UserDefaultStore: BaseUserDefaultStore {
         }
     }
 }
+
+extension UserDefaultStore {
+    private func readValuesFrom(_ userDefaults: UserDefaults) {
+        for key in LocalUserDefaultKey.allValues {
```
linuxwolf: This is effectively moving everything from the "old" UserDefaults to the shared one, yes?

Is there a point in keeping the `LocalUserDefaultKey` separate from `UserDefaultKey`?

sashei: The general idea is to keep settings that aren't relevant to the credential provider isolated to the application. For settings like `preferredBrowser` etc, we don't need to include the`Setting.PreferredBrowser` class in the credential provider. The impetus for keeping the keys separated as well is the `defaultValue` parameter required for all `UserDefaultKey` values.

PR #666, opened 2018-08-30
locking and unlocking, providing entries from quicktype bar

### LINKED ISSUE #610: Locking & Unlocking
*** Should be completable with dummy data ***

### Acceptance Criteria

Display TouchID / FaceID prompt when relevant while:
- [ ] `[1]` attempting to use a credential from the QuickType bar while the app is locked (dev note: `provideCredentialWithoutUserInteraction`)
- [ ] `[2]` attempting to access the credential list to select a credential (dev note: `prepareCredentialList`)

The outcome for the above should be as follows:
- [ ] `[1]` provide credential to QuickType requesting app
- [ ] `[2]` display list of credentials with pre-populated filter field

### LINKED ISSUE #615: Provide entry when requested via the QuickType bar
When a user selects an entry via the QuickType bar, provide the full entry to populate the text field.

Don't forget to `touch` the item so that its use is reflected in `Recently Used`

NOTE: this story assumes that the application is unlocked. Locking / unlocking is handled in #610.

depends on #612

sashei:
resolves #610 
resolves #615 

** testing notes **
There are three cases in which we expect to see autolock behavior manifest (i.e., prompted for device authentication when the appropriate amount of time per user setting has elapsed):

- the regular way in `Lockbox.app`
- by going to the list of credentials from the QuickType bar (selecting `Lockbox...` from the action sheet)
- enabling Lockbox from Settings as a credential provider

You will note that accessing the credentials by selecting them DIRECTLY from the QuickType bar is not included in this list; this is because iOS prompts users for biometric authentication before doing the autofill and we felt that prompting them from Lockbox as well was overkill / annoying. So in the case of autofilling from the QuickType bar, this is the expected behavior:
- if the app was locked, it will stay locked
- if the app was unlocked, the timer will not be reset, so if, eg., a user has the timer set to 30 minutes, autofills from QuickType bar 15 minutes after use, the app will still lock 15 minutes later

### DIFF
--- a/Shared/Store/BaseDataStore.swift
+++ b/Shared/Store/BaseDataStore.swift
@@ -159,6 +159,7 @@ class BaseDataStore {
                         self.profile.syncManager?.applicationDidEnterBackground()
                     case .foreground:
                         self.profile.syncManager?.applicationDidBecomeActive()
+                        self.handleLock()
                     case .upgrade(let previous, _):
                         if previous <= 2 {
                             self.handleLock()
@@ -413,13 +414,13 @@ extension BaseDataStore {
             return
         }
         self.syncSubject.onNext(.ReadyToSync)
+
+        // default to locked state on initialized
+        self.storageStateSubject.onNext(.Locked)
         self.handleLock()
     }
 
     internal func handleLock() {
-        // default to locked state
-        self.storageStateSubject.onNext(.Locked)
-
         self.userDefaults.onAutoLockTime
                 .take(1)
                 .subscribe(onNext: { autoLockSetting in
--- a/Shared/Store/BaseDataStore.swift
+++ b/Shared/Store/BaseDataStore.swift
@@ -94,7 +94,7 @@ class BaseDataStore {
     private let keychainWrapper: KeychainWrapper
     private let userDefaults: UserDefaults
     internal var profile: FxAUtils.Profile
-    private let dispatcher: Dispatcher
+    internal let dispatcher: Dispatcher
 
     public var list: Observable<[Login]> {
         return self.listSubject.asObservable()
@@ -190,6 +190,11 @@ class BaseDataStore {
                 .disposed(by: self.disposeBag)
 
         self.setInitialState()
+        self.initialized()
+    }
+    
+    public func initialized() {
+        fatalError("not implemented!")
     }
 
     public func get(_ id: String) -> Observable<Login?> {
@@ -411,7 +416,7 @@ extension BaseDataStore {
         self.handleLock()
     }
 
-    private func handleLock() {
+    internal func handleLock() {
         // default to locked state
         self.storageStateSubject.onNext(.Locked)
 
--- a/CredentialProvider/Action/CredentialStatusAction.swift
+++ b/CredentialProvider/Action/CredentialStatusAction.swift
@@ -6,7 +6,7 @@ import Foundation
 import Storage
 
 enum CredentialStatusAction: Action {
-    case extensionConfigured, userCancelled, loginSelected(login: Login, relock: Bool)
+    case extensionConfigured, userCanceled, loginSelected(login: Login, relock: Bool)
 }
 
 extension CredentialStatusAction: Equatable {
@@ -14,7 +14,7 @@ extension CredentialStatusAction: Equatable {
         switch (lhs, rhs) {
         case (.extensionConfigured, .extensionConfigured):
             return true
-        case (.userCancelled, .userCancelled):
+        case (.userCanceled, .userCanceled):
             return true
         case (.loginSelected(let lhLogin, let lhRelock), .loginSelected(let rhLogin, let rhRelock)):
             return lhLogin == rhLogin && lhRelock == rhRelock
--- a/CredentialProvider/Action/CredentialStatusAction.swift
+++ b/CredentialProvider/Action/CredentialStatusAction.swift
@@ -5,5 +5,5 @@
 import Foundation
 
 enum CredentialStatusAction: Action {
-    case extensionConfigured
+    case extensionConfigured, userCancelled
 }
--- a/CredentialProvider/Action/CredentialStatusAction.swift
+++ b/CredentialProvider/Action/CredentialStatusAction.swift
@@ -6,5 +6,5 @@ import Foundation
 import Storage
 
 enum CredentialStatusAction: Action {
-    case extensionConfigured, userCancelled, loginSelected(login: Login)
+    case extensionConfigured, userCancelled, loginSelected(login: Login, relock: Bool)
 }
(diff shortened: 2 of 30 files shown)


devinreams:
@sashei Holy cow I installed the latest build and this just worked for me. So. Awesome. 🥇

sashei:
@nickbrandt ohhh yes!!! we made a UX decision yesterday around the QuickType bar specifically but I'm realizing that it's not documented here. I know it's EOD in Denver, so I will make sure to update the description of this PR & the attendant issues for you by tomorrow :)

devinreams:
>You will note that accessing the credentials by selecting them DIRECTLY from the QuickType bar is not included in this list; this is because iOS prompts users for biometric authentication before doing the autofill and we felt that prompting them from Lockbox as well was overkill / annoying

Extra context: I think one primary case Apple had in mind for this extra app-provided authentication step during QuickType tap/selection (as I understood it from the videos and docs) is for if those password managers that check their database and saw the user needed to enter their "master password" to unlock the database, from a technical perspective, in order to get the un-encrypted password.

Just because the login (website/app and username) is shown on the QuickType bar, from a security perspective the password is still not yet available to the app until the user explicitly taps and agrees at which point the app itself may not have the password available until the user takes _one more_ step (enter a master passwords).

devinreams:
@sashei I hit a scenario where I was given a dialog: "**Credential list not available yet** Please check back later"

- opened Lockbox, manually locked
- opened Safari and browsed to reddit.com to log in
- was shown QT for my login and tapped it (Face ID and then inserted ✅)
- closed Safari and opened Lockbox (it prompted for Face ID and opened)
- opened Safari and opened reddit.com to log in again
- no QT shown for login but just "Passwords" (and the key icon)
- tapped "Passwords" and Lockbox slides over, prompts for Face ID, then throws the dialog ❌ 
- switched to and opened Lockbox again
- switched to Safari again and it shows me the QT entry and tapping it fills the form ✅ 

Any idea if the "Credential list not available yet" is related to these changes? Is it something you've seen, too?

nickbrandt:
@sashei thanks for clarifying the lock cases. Tested again with following results:

Downloaded build > signed in > confirmed > dropped in entry list
Enabled Lockbox in AutoFill
Went to twitter.com in Safari
No quicktypebar suggestions shown ❌ 
Went back to Lockbox, pulled to initiate sync
Went back to twitter.com in Safari > suggestions now showing ✅ 
Tapped entry, presented in context auth, authenticated, filled ✅ 
Went back to Lockbox, manually Locked
Went back to twitter.com in Safari
Tapped key icon > selected “Lockbox…” > Lockbox slides up (halfway) and prompt displays over top (photo) ⚠️ 
Hit Cancel instead of authenticating > displays blank Lockbox screen (screenshot) ❌ 
Hit Home and went back to Safari > Tapped key > no auth required > displays Credential list not available yet dialog (assuming this dialog is your placeholder for entry list) ❌ 

slide
blank

sashei:
I've seen some funny behavior with QTB not being populated as well, let me dig into that y'all! Thank you for the detailed steps to repro, and I'll circle back around when I've got something in place :)

sashei:
Also, the `Credential List not available yet` is a placeholder I put in for this story, so that's expected to see in whatever situation you'd expect to see the credential list.

nickbrandt:
I tried doing a full re-install as @devinreams suggested (same as his initial steps above) to see if it changed anything on my end.

Manually locked app > Went to AutoFill settings and tapped to enable Lockbox > Touch ID prompt ✅  (note: this is having the same issue I outlined earlier with the prompt going to the Cred List, where Lockbox hasnt fully slid in from bottom - photo attached) ❌ 
If I cancel out of that prompt instead of authenticating, I still have the blank Lockbox screen presented to me and have to force quit out of Settings to get rid of it. ❌ 

Went ahead and enabled Lockbox (authenticated) > Loading screen > Enabled ✅ 
Went to Safari and opened twitter.com > QuicktypeBar suggestions showing > Select suggestion > Prompt to unlock > Fills ✅  (if choose Cancel, goes Prompt dismisses and remains in previous state ✅ 

Tried route of tapping key icon instead > selected "Lockbox..." (within 5 min of authenticating for enabling in Settings) > no prompt > Entry List placeholder slides up ✅ 

Went back to Lockbox app > Still locked ❌ ? > Unlocked with biometrics > manually Lock Now > go back to route of tapping key icon in Safari > prompted with biometrics  > authenticate > Entry List placeholder slides up ✅ 
(note: this has same issue as outlined above in Settings when prompted, and when cancelling gets stuck in blank Lockbox state ❌ )

Tested these scenarios in Twitter app as well, with same results. Though sometimes I did not see the QuicktypeBar suggestions (typically after canceling AutoFill, leaving sign in page, and returning)

I'm still a little unclear on the relationship between lock states in our Lockbox app vs. using AutoFill. It seems that once I authenticate (either when enabling in Settings or going to view the Credential List from QuicktypeBar) that the timer resets and doesn't require authentication for AutoFill tasks, but the Lockbox app remains locked. Is that the intended behavior? It can be confusing, as I will not need to authenticate at times, but when I go to the Lockbox app it is locked.

sashei:
@nickbrandt I don't know what the best way is to handle cancel / failure from authenticating, I think I've addressed everything else that you put in here. What's the desired behavior there?

nickbrandt:
@sashei well, for enabling AutoFIll, the **desired** behavior would be to not enable (if we are requiring authentication), though I know we cannot do that.

As long as canceling out of one of those instances simply brings the user back to their last spot (Settings, App, Safari), then I think we are good. The issue I was having is that if I choose to Cancel one of those auth requests, I get put into a blank Lockbox app screen (screenshot above), and have to force quit out of whatever I was in (Settings/Safari/App).

sashei:
@nickbrandt yes, so after a conversation with @linuxwolf about locking / unlocking scenarios, we decided that it was best to request an unlock in order to populate the QuickType bar, even if the app is still enabled when the user cancels. I've addressed the desired cancel behavior (I was curious if you'd prefer another button there, but I think cancelling the screen entirely is simpler / better UX so 👍🏼) and will push when I am in the office.

Shared/Store/BaseDataStore.swift
```diff
@@ -421,24 +421,14 @@ extension BaseDataStore {
...
-                            self.lock()
-                        }
-                    }
-                })
-                .disposed(by: self.disposeBag)
+        let date = NSDate(
+                timeIntervalSince1970: self.userDefaults.double(
+                        forKey: UserDefaultKey.autoLockTimerDate.rawValue))
+
+        if date.timeIntervalSince1970 > 0 && date.timeIntervalSinceNow > 0 {
```
linuxwolf: checks that are not absolute time make me nervous, but this seems correct ...

sashei: this code has not changed since the app launched; I'm just removing the special case for `Never` :p

devinreams: I hold myself responsible for side effects here...

PR #668, opened 2018-08-30
Add "Add a browser" documentation

### LINKED ISSUE #546: Add documentation for developers to add their browser
In order to support as many browsers — and user choice of browser — as possible, we should document how developers can add their browsers to the list we provide in Settings.

In Firefox-iOS we do something similar with password managers: https://github.com/mozilla-mobile/firefox-ios/blob/master/Client/Frontend/Share/ShareExtensionHelper.swift#L105

devinreams:
Fixes #546 

This adds a new documentation page based on @joeyg's example branch:
https://github.com/mozilla-lockbox/lockbox-ios/compare/master...joeyg:duckduckgo?expand=1

I think I technically worded this doc correctly but would appreciate a review from @joeyg or @mozilla-lockbox/mobile-engineering.

Note: the readthedocs theme doesn't do a great job with the code formatting:

screen shot 2018-08-30 at 3 34 56 pm

...but it looks great in rendered markdown (which folks can get to from this repo, or from the website clicking the 'Edit in GitHub' link):

screen shot 2018-08-30 at 3 36 57 pm

So I think its fine as it is but open to easy wins before merging.

### DIFF
--- a/docs/add-a-browser.md
+++ b/docs/add-a-browser.md
@@ -1,0 +1,49 @@
+# Add a browser setting to Firefox Lockbox
+
+_These instructions are for developers wanting to add another browser setting so when users tap to open an entry's URL it will be sent to the web browser of their choice._
+
+
+1. Find and add the query scheme the new browser has registered to `LSApplicationQueriesSchemes` in `Common/Resources/Info.plist`. For example we'll use [DuckDuckGo](https://github.com/mozilla-lockbox/lockbox-ios/compare/master...joeyg:duckduckgo?expand=1):
+
+  ```
+  <string>ddgQuickLink</string>
+  ```
+
+
+2. Then add a new case to `PreferredBrowserSetting` in `ExternalLinkAction.swift`:
+
+  ```swift
+  case DuckDuckGo
+  ```
+
+3. Define that new case in the `getPreferredBrowserDeeplink` switch statement and return the string (with that new query scheme as the expected URL syntax):
+
+  ```swift
+  case .DuckDuckGo:
+    return URL(string: "ddgQuickLink://\(url)")
+  ```
+
+3. Add a constant (string) for the name of the browser (this will be used for the setting) to `Common/Resources/Constants.swift`:
+
+  ```swift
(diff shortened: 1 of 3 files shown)


docs/add-a-browser.md
```diff
@@ -0,0 +1,49 @@
...
+  ```
+
+
+2. Then add a new case to `PreferredBrowserSetting` in `ExternalLinkAction.swift`:
+
+  ```swift
+  case DuckDuckGo
+  ```
+
+3. Define that new case in the `getPreferredBrowserDeeplink` switch statement and return the string (with that new query scheme as the expected URL syntax):
```
sashei: developers might need to URL-encode the URL that they're sending here. I know there's some trickiness here with how Firefox for iOS, at least, expects to receive URLs, so maybe include a note about "test deeplinks to make sure that they open as expected" and URL-encoding...

docs/add-a-browser.md
```diff
@@ -0,0 +1,49 @@
...
+    return URL(string: "ddgQuickLink://\(url)")
+  ```
+
+3. Add a constant (string) for the name of the browser (this will be used for the setting) to `Common/Resources/Constants.swift`:
+
+  ```swift
+  static let settingsBrowserDuckDuckGo = NSLocalizedString("settings.browser.duckduckgo", value: "Duck Duck Go", comment: "Duck Duck Go Browser")
+  ```
+
+4. Then include the constant (name of the browser) to the case in the `toString()` function back in `Action/ExternalLinkAction.swift`:
```
sashei: nit: This is a useful step to have, but I think that Swift enums will cause a build error to happen if developers _don't_ do this... The steps are nice to have too :)

docs/add-a-browser.md
```diff
@@ -0,0 +1,49 @@
...
+
+  ```swift
+  case .DuckDuckGo:
+    return Constant.string.settingsBrowserDuckDuckGo
+```
+
+5. Also add the name constant to the `initialSettings` variable in `Presenter/PreferredBrowserSettingPresenter.swift` so the `PreferredBrowserSettingPresenter` class can mark the browser as "checked" when selected:
+
+  ```swift
+  lazy var initialSettings = [
```
sashei: 😬 wish this was an enum too but I suppose that's out of scope for this PR!!

PR #673, opened 2018-08-31
Autofill Display disclaimer

### LINKED ISSUE #611: Display disclaimer if user enables Lockbox as autofill credential provider without authenticating
### Acceptance Criteria

Install Lockbox without logging in
Enable Lockbox as a credential provider from Settings

Behavior: Extension displays a disclaimer directing users to authenticate in Lockbox.app

### Visual Design

**InVision**
https://mozilla.invisionapp.com/share/BQEUSO0S2#/313162524_AutoFill_-_Autofill_Passwords_-_Enable_-_Not_Authenticated

**Zeplin**
https://zpl.io/VKqWZP8

Note: "Sign In" button would navigate users to the FxA sign in flow within the Lockbox app.
If we go the low level of effort route, and not direct the user to our app, we would need to update the copy on the button to "OK" instead of "Sign In".

depends on #612

joeyg:
Resolves #611.

simulator screen shot - iphone 8 plus - 2018-08-31 at 20 36 04

linktoapp

### DIFF
--- a/CredentialProvider/Presenter/CredentialProviderPresenter.swift
+++ b/CredentialProvider/Presenter/CredentialProviderPresenter.swift
@@ -44,6 +44,8 @@ class CredentialProviderPresenter {
                 switch action {
                 case .extensionConfigured:
                     self.view?.extensionContext.completeExtensionConfigurationRequest()
+                case .cancel:
+                    self.view?.extensionContext.cancelRequest(withError: NSError(domain: ASExtensionErrorDomain, code:ASExtensionError.userInteractionRequired.rawValue))
                 }
             })
             .disposed(by: self.disposeBag)
--- a/CredentialProvider/Presenter/CredentialProviderPresenter.swift
+++ b/CredentialProvider/Presenter/CredentialProviderPresenter.swift
@@ -44,8 +44,6 @@ class CredentialProviderPresenter {
                 switch action {
                 case .extensionConfigured:
                     self.view?.extensionContext.completeExtensionConfigurationRequest()
-                case .cancel:
-                    self.view?.extensionContext.cancelRequest(withError: NSError(domain: ASExtensionErrorDomain, code:ASExtensionError.userInteractionRequired.rawValue))
                 }
             })
             .disposed(by: self.disposeBag)
--- a/CredentialProvider/Presenter/CredentialWelcomePresenter.swift
+++ b/CredentialProvider/Presenter/CredentialWelcomePresenter.swift
@@ -18,11 +18,7 @@ class CredentialWelcomePresenter: BaseWelcomePresenter {
 
     private var okButtonObserver: AnyObserver<Void> {
         return Binder(self) { target, _ in
-            if let url = URL(string: "lockbox://fxa") {
-                UIApplication.shared.open(url, options: [:], completionHandler: nil)
-            }
-
-            self.dispatcher.dispatch(action: CredentialStatusAction.cancel)
+            self.dispatcher.dispatch(action: CredentialStatusAction.extensionConfigured)
         }.asObserver()
     }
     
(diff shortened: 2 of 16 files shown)


devinreams:
That's pretty slick @joeyg.

Is it easy to do drop straight into the FxAWebView? I believe the "upgrade" prompt for the 1.2 OAuth upgrade does...

sashei:
@joeyg I'd think push on top of the Welcome view.

joeyg:
@sashei @devinreams a more robust deeplinking system is probably worth exploring but I'll let you decide if this is the time for that. For now it opens up the FxA screen.

CredentialProvider/Presenter/CredentialWelcomePresenter.swift
```diff
@@ -14,6 +15,16 @@ class CredentialWelcomePresenter: BaseWelcomePresenter {
     }
     
     private let credentialProviderStore: CredentialProviderStore
+
+    private var okButtonObserver: AnyObserver<Void> {
+        return Binder(self) { target, _ in
+            if let url = URL(string: "lockbox://fxa") {
```
sashei: This is not a secure way to do universal linking; someone could register for this domain and phish FxA logins.

sashei: Rather than deeplinking here, let's just dispatch the configured action and let users take the action on their own.

CredentialProvider/Presenter/CredentialProviderPresenter.swift
```diff
@@ -44,6 +44,8 @@ class CredentialProviderPresenter {
                 switch action {
                 case .extensionConfigured:
                     self.view?.extensionContext.completeExtensionConfigurationRequest()
+                case .cancel:
```
sashei: Have you tested that you can cancel this request and it stops the user from enabling autofill? I could be mistaken but I thought that `cancelRequest` was only for credential requests.

lockbox-ios/Common/AppDelegate.swift
```diff
@@ -71,6 +71,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         return true
     }
 
+    func application(_ application: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
```
sashei: See above comment on deeplinking.

devinreams:
Sorry @joeyg, I asked about the deeplink without realizing I was asking for something more than what we should probably sign up for here. Thanks for the help. Sorry for the unintented run-around. 😔

CredentialProvider/Presenter/CredentialProviderPresenter.swift
joeyg: Yeah it doesn't seem to do what I wanted.

nickbrandt:
The dialog looks good, only a small addition of a period (".") at the end of the last sentence.

@sashei @devinreams I have a question about how this functions.
From my initial understanding of this, we were unable to enable AutoFill for our app without it first being authenticated (signed in). From how this currently works in the build, the user is displayed with the dialog about requiring to sign in, but Lockbox still gets enabled. This would leave me to believe that the user does not need to come back and enable Lockbox in AutoFill once they have signed in. 

Is the issue more about the credential store not being populated (and empty) until they sign in? If that is the case, then we would probably need to update this copy a bit. Also, is there, or do we need to create an issue to account for the case where someone has not signed in yet, and chooses "Lockbox" from the action sheet?
action-sheet

sashei:
@nickbrandt yes, your understanding is correct. There is not a way to "disallow" the system from enabling Lockbox even if the user is not authenticated. However, we won't populate the credential store until they have done so, and they don't need to enable Lockbox a second time from settings -- I believe that it's already a part of the onboarding story to not display the `Enable Autofill` screen if Lockbox is already enabled as a credential provider.

Good catch on the UI in the action sheet situation; I think with our current implementation, users will likely just see an empty list when selecting Lockbox from the action sheet without authenticating first. It might be useful to show something more informative there though....

nickbrandt:
@sashei gotcha. I was asking about needing to go back to Settings because we currently have that in the dialog copy. We should probably update that to something like the attached image, if that makes sense to you. I'll add this to the document we sent Michelle for additional copy requests, but don't want to hold up this PR.

screen shot 2018-09-05 at 3 04 43 pm

PR #674, opened 2018-09-01
autofill authentication

### LINKED ISSUE #613: Populate `ASCredentialIdentityStore` after authenticating if Lockbox is already enabled as an autofill credential provider
After authentication, populate the `ASCredentialIdentityStore.shared`.

### Acceptance Criteria

- Install Lockbox
- Enable Lockbox as a credential provider in Settings
- Authenticate

Expected: Credentials are displayed in the QuickType bar

joeyg:
Resolves #613. Not sure this is the full scope of the ticket - feel free to ignore this PR if it isn't helpful.

### DIFF
--- a/CredentialProvider/Presenter/CredentialWelcomePresenter.swift
+++ b/CredentialProvider/Presenter/CredentialWelcomePresenter.swift
@@ -33,6 +33,12 @@ class CredentialWelcomePresenter: BaseWelcomePresenter {
     }
 
     override func onViewReady() {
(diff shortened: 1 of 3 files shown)


sashei:
@joeyg thanks for the PR -- you're right that this doesn't quite address the intention of #613. This case is being handled in PR #666; rather, the code for #613 should live in `Lockbox.app`, the idea being that we can't populated the `ASCredentialStore` in the settings if we haven't authenticated yet. So after a user authenticates in `Lockbox.app`, we can just dispatch `CredentialProviderAction.refresh`.

Issue #512, opened 2018-06-27
Dynamic font size improvements


devinreams:
Follow up to #491 #470

- [x] Confused about why icons are behaving in that manner, as well as truncation acting two different ways https://github.com/mozilla-lockbox/lockbox-ios/pull/628
  <img width="376" alt="screen shot 2018-06-27 at 9 37 58 am" src="https://user-images.githubusercontent.com/6164801/41984529-24db75e2-79ee-11e8-99ee-19fd9e39cc71.png">

- [x] "Lock Now" button not getting bigger relative to font size, also losing padding from bottom
  <img width="377" alt="screen shot 2018-06-27 at 9 38 23 am" src="https://user-images.githubusercontent.com/6164801/41984599-45dfe3a4-79ee-11e8-81c1-ac3d01d370ec.png">

- [ ] Why is search truncating so early, relative to the space available?
  <img width="376" alt="screen shot 2018-06-27 at 9 38 36 am" src="https://user-images.githubusercontent.com/6164801/41984667-734a77d2-79ee-11e8-887e-1a738d1c6a16.png">

- [ ] "Learn how to edit this entry" appears to be fixed to the base of the screen, overlapping with the rest of the content on the detail... 
  <img width="393" alt="screen shot 2018-06-27 at 9 38 48 am" src="https://user-images.githubusercontent.com/6164801/41984687-84e5d9fa-79ee-11e8-8098-dc59c8fa1ac5.png">

- [ ] Unlock prompt getting cut off on the bottom
  <img width="390" alt="screen shot 2018-06-27 at 9 44 51 am" src="https://user-images.githubusercontent.com/6164801/41984834-e3e0edbe-79ee-11e8-90da-61c500160b4a.png">

- [ ] I noticed the second bullet point on the onboarding confirmation screen (encryption) was not dynamically sized. It appears to be a text view, not a label like the other two. Should that be swapped out?

sigito:
Overall Apple doesn't scale navigation icons, title, and search bars. Should lockbox also not try to scale those icons?

changecourse:
Agreed @sigito, I think we'd want to lock those icons down so they aren't adversely affecting the layout, and are in-line with Apple's heuristics.

PR #675, opened 2018-09-03
Size Lock Now button for dynamic type

sigito:
Address "Lock Now" button scaling issues in #512. Connected to #512

### Footer view sizing
Apparently, TableView's footer view would not resize for the internal content size changes. Thus we have to manually trigger its resize. See: https://collindonnell.com/2015/09/29/dynamically-sized-table-view-header-or-footer-using-auto-layout/ and https://lists.apple.com/archives/cocoa-dev/2014/Jun/msg00127.html.

### Borders
Borders layers are not resized with the view. Thus if we scale the view, the bottom border would still be drawn at the old position, now in the middle of the view. For this we have two options:
1. Keep a reference to added layers, and layout them together with the view.
1. Display a border as part of the view layer, which is automatically sized.
Since the footer is full width, it is safe to simply set the border on the view.

### Results
| Minimum font size | Maximum font size |
| ------------------ | ------------------- |
| min | max |

### DIFF
--- a/lockbox-ios/Presenter/SettingListPresenter.swift
+++ b/lockbox-ios/Presenter/SettingListPresenter.swift
@@ -10,7 +10,7 @@ import LocalAuthentication
 
 protocol SettingListViewProtocol: class, AlertControllerView {
     func bind(items: Driver<[SettingSectionModel]>)
-    var onSignOut: ControlEvent<Void> { get }
+    var onLockNow: ControlEvent<Void> { get }
 }
 
 class SettingListPresenter {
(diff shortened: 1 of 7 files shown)


lockbox-ios/Presenter/SettingListPresenter.swift
```diff
@@ -113,7 +113,7 @@ class SettingListPresenter {
 
         self.view?.bind(items: settingsConfigDriver)
 
-        self.view?.onSignOut
+        self.view?.onLockNow
```
sashei: 😍

PR #679, opened 2018-09-05
Autofill onboarding

### LINKED ISSUE #617: Onboarding
**Product question**: If users have already enabled Lockbox as an autofill credential provider, should we display this section of the onboarding flow?

Onboarding screen is displayed after authentication with a button to take user to relevant settings and Not Now button to dismiss

Onboarding Flow to Settings
https://mozilla.invisionapp.com/share/54E2A6ZKF#/304302929_AutoFill_-_Onboarding

joeyg:
Resolves #617.

TODO:
 * Test link to settings screen

iPhone 8:
simulator screen shot - iphone 8 - 2018-09-05 at 15 45 50

iPhone 8 plus:
simulator screen shot - iphone 8 plus - 2018-09-05 at 15 43 32


devinreams:
For assets and sizes they _should_ now be available in Zeplin: https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5b8434bf6712a70856032f61

As I was confirming myself I figured I'd drop in what I saw here:

- "AutoFill your logins..." heading should be whatever dynamic iOS font size is closest to 24pt and SFUIText-Light (rgb = 0 0 0)
- "No longer are the days.." blurb should be whatever dynamic iOS font size is closest to 15pt and SFUIText-Regular (rgb = 12 12 13)
- "Set Up AutoFill" button and "Not Now" button closest to 15pt

image
image 2x
image 3x

nickbrandt:
@joeyg that is correct. This image shows the mappings we had created when changing to dynamic text. I think the only font size that changed for the default size is the heading (AutoFill your logins...), which I need to update in the designs. 

screen shot 2018-09-05 at 10 53 37 am

Added those mappings for reference here:
https://zpl.io/bWLJJdM
https://zpl.io/Vknvvz5

lockbox-ios/Presenter/FxAPresenter.swift
```diff
@@ -52,7 +52,15 @@ class FxAPresenter {
 extension FxAPresenter {
     func matchingRedirectURLReceived(_ navigationURL: URL) {
         self.dispatcher.dispatch(action: OnboardingStatusAction(onboardingInProgress: true))
-        self.dispatcher.dispatch(action: LoginRouteAction.onboardingConfirmation)
+        self.dispatcher.dispatch(action: self.getOnboardingRoute())
         self.dispatcher.dispatch(action: AccountAction.oauthRedirect(url: navigationURL))
     }
+
+    private func getOnboardingRoute() -> LoginRouteAction {
```
sashei: nice!

sashei: I think the other piece here is checking whether autofill is already enabled or not.. if the user has already enabled autofill, then we shouldn't display the autofill onboarding screen.

joeyg: @sashei good catch! Because getting `CredentialProviderStore.state` is an observable should there be a `Observable<LoginRouteAction>` property that holds the state of what action to send?

sashei: Hm, I suppose it has to be async in some way... maybe have a function called `beginOnboarding()` and that can `.take(1)` from the `CredentialProviderStore.state` to dispatch the appropriate onboarding route action?

I realize looking at the `CredentialProviderStore` that there is no state for `.Allowed` so you might need to add that.

joeyg:
I resolved the test issues. There is still the outstanding issue if you comeback to the app from Settings you are dropped in the Listview instead of the final Onboarding screen. I suspect that has been an issue since https://github.com/mozilla-lockbox/lockbox-ios/commit/50879d31302623c3cd1c4100e19b23f7d59d3d44

lockbox-ios/Storyboard/AutofillOnboarding.storyboard
```diff
@@ -0,0 +1,91 @@
...
+                                <rect key="frame" x="40" y="364" width="295" height="54"/>
+                                <fontDescription key="fontDescription" type="system" pointSize="15"/>
+                                <color key="textColor" red="0.047058823529411764" green="0.047058823529411764" blue="0.050980392156862744" alpha="0.84999999999999998" colorSpace="calibratedRGB"/>
+                                <nil key="highlightedColor"/>
+                            </label>
+                            <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="AutoFIll your logins right from your browser or app." textAlignment="center" lineBreakMode="tailTruncation" numberOfLines="2" baselineAdjustment="alignBaselines" adjustsFontForContentSizeCategory="YES" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="Krc-YI-yKW">
```
devinreams: I spot a capital "I" in "FIll" here. 👀

lockbox-ios/Presenter/FxAPresenter.swift
```diff
@@ -22,6 +22,23 @@ class FxAPresenter {
     fileprivate let dispatcher: Dispatcher
     fileprivate let accountStore: AccountStore
 
+    fileprivate var _credentialProviderStore: Any?
+
+    @available (iOS 12, *)
+    private var credentialProviderStore: CredentialProviderStore? {
```
sashei: It doesn't look like this will ever return nil, why is it optional?

lockbox-ios/Presenter/FxAPresenter.swift
```diff
@@ -39,20 +56,50 @@ class FxAPresenter {
         self.accountStore = accountStore
     }
 
+    @available(iOS 12, *)
```
sashei: Is there a corresponding init that's available not on iOS 12?

devinreams:
>Returned to app, has skipped over "Welcome to Lockbox" screen ❌ (sounds as if part of another issue)

I added a new issue so we don't lose that point: https://github.com/mozilla-lockbox/lockbox-ios/issues/688

lockbox-ios/Presenter/FxAPresenter.swift
joeyg: yep ln50 has a init that is for other versions.

lockbox-ios/Presenter/FxAPresenter.swift
joeyg: good catch!

PR #685, opened 2018-09-06
Fix Safari UI test (3)

devinreams:
After talking to @isabelrios I wanted to quickly add this to the autofill branch so we can "get green" and confident everything is still working as expected before we wrap it up.

By waiting for the existing of the buttons, even if the page takes a while to load, the test should continue and pass much more consistently now.


devinreams:
This is going into the bigger improvements PR here:
https://github.com/mozilla-lockbox/lockbox-ios/pull/687/files#diff-84e408a636b68acfd8ec676341dec30aR119

PR #691, opened 2018-09-08
Launch biometrics if logged in but locked

### LINKED ISSUE #683: Display Touch ID/Face ID/Passcode prompt on cold open of app
Currently, when a user foregrounds the app when it is locked (after already being signed in), and they have biometrics or passcode setup on their device, we display the prompt over top the lock screen so they do not need to initiate it themselves (tap the "Unlock" button). We would like to extend this behavior to also happen on cold open of the app (if a user has already signed in).

joeyg:
Resolves #683.


sashei:
@joeyg I haven't tested this, but one possible gotcha is that we do not want to launch biometric auth after someone hits `Lock Now`, this should only happen from cold launch.

joeyg:
good catch @sashei - i'll re-work this PR.

devinreams:
@joeyg sorry for any hassle but would you mind rebasing against `master` and changing this PR to compare accordingly? autofill is now complete and merged! 🔀

jhugman:
Where are we up to with this PR? Could we close this, or rebase onto master?

joeyg:
@jhugman I put this PR on pause because of the issue @sashei mentioned above. I think it might make sense to fix this at the same time as #693.

Issue #624, opened 2018-08-11
Capture autofill telemetry events


sandysage:
As the product owner, I want to know:

- is this feature discoverable and can people set it up (enabling it)?
- are people are using this feature (filling a credential)?
- how often do we have a matching result in the quicktype bar versus the user going to the Lockbox list view?
- where are they dropping off? can people complete the task of logging in?

Are just a few I can think of off the top of my head. @lrrationalagent, I opened this issue so that we:

- [ ] identify any additional questions we want to answer with this feature
- [ ] capture the additional events here and in our metrics.md
- [ ] add the events

irrationalagent:
We'd like, at a minimum, to have events for the following actions:
- [ ] autofill enabled in app (if there is more than one way to do this e.g. thru onboarding or thru settings, its OK to have a generic event that covers all touchpoints. if its not extra work to have a unique event per touchpoint, then let's do that)
- [ ] telemetry event corresponding to `provideCredentialWithoutUserInteraction` (credential provided directly to quicktype
- [ ] telemetry event corresponding to `prepareCredentialList` (credential list displayed)
- [ ] some event corresponding to a successful autofill. lmk if this is hard - I think its relatively important to have since it allows us to know, in combination with two events prior, the proportion of people initiating the autofill process that actually get to the desired end-state. but i get that this may not be possible on our end (i.e. the actual filling is done outside of lockbox, and we don't get any sort of callback)

nice to have if easy to do:
- [ ] for `prepareCredentialList` related events, pass the number of entries in the filtered list to the `value` field of the event
- [ ] prompt to unlock displayed when trying to autofill
- [ ] unlock successful/unsuccessful when trying to autofill

if we had the lock/unlock events we could know for sure if people are getting tripped up at that step, if it happens that the number of successful autofills is much smaller than the number that are initiated. if we don't have those events, then we will just have to infer that it might be an issue with the locking without knowing 100%.

PR #702, opened 2018-09-12
autofill telemetry

irrationalagent:
resolves https://github.com/mozilla-lockbox/lockbox-ios/issues/624

So I pulled down master after the autofill merge,  added @sashei's code that I had stashed from our pairing on monday, resolved any conflicts, added the actual events, and one unit test.

Currently, I'm not able to verify that this actually works - usually, i add the following print statements to `TelemetryAction.swift` so that the event properties are logged to the console when the events are triggered:

```swift
    lazy var telemetryActionListener: AnyObserver<TelemetryAction> = {
            target.telemetry.recordEvent(
                    value: action.value,
                    extras: action.extras
            )
            print (    // I added this print statement
              TelemetryEventCategory.action.rawValue,
              action.eventMethod.rawValue,
              action.eventObject.rawValue,
              action.value,
              action.extras
            );
        }.asObserver()
    }()
}
```
however doing so for the new events does not lead to them getting console logged when they should. but: the one unit test that I added to `CredentialStatusActionSpec.swift` to check that the proper event methods are being passed does work, so I'm scratching my head a bit. either

(1) the new events in the extension don't pass thru the code above for some reason that i obviously don't get
(2) its not working.

I have a feeling that its the latter. Any pointers would be welcome


sashei:
@irrationalagent I haven't reviewed this yet, but to see logs for the credential provider, you have to attach to it specifically in Xcode. You won't see them just by running normally.

To attach to the credential provider, navigate (in Xcode) to `Debug -> Attach to Process by Name or PID` and type in `CredentialProvider`. Then, when you launch the credential provider, the logs will show up when selecting CredentialProvider under this menu on the left bar:
screen shot 2018-09-12 at 8 02 32 am

Sorry this process is kind of lame! But that would be why they aren't showing up :)

PR #727, opened 2018-09-14
Set disclosure indicator if not in the AutoFill extension

### LINKED ISSUE #725: Hide disclosure accessory (chevrons) on AutoFill entry list
Spun out of #712. A leftover todo after #616 and #682: The chevrons on the right side of each row are not necessary (since no detail to drill into).

I see the Base/ItemList/CredentialItemList views but was not smart enough to properly check/override the accessoryType accordingly. 😔

devinreams:
...and removed as a default accessory shown in the Interface Builder.

Fixes #725 

Now there are no chevron ">" arrows in the autofill entry list extension.

Too easy? Better way to do this?


Shared/View/BaseItemListView.swift
```diff
@@ -160,6 +160,10 @@ extension BaseItemListView {
 
                         cell.titleLabel.text = title
                         cell.detailLabel.text = username
+                        
+                        if (self.extensionContext == nil) {
+                            cell.accessoryType = .disclosureIndicator
+                        }
```
jhugman: This doesn't drive me wild with approval. I'd prefer to use a method that gets overridden in the subclass (removes the check for `extensionContext` but adds a method call).

However, given timescales, I think I'd label this as #technical-debt, approve it and move on.

devinreams: Agreed on both points. I'll file a tech-debt story to come back to this since I'm not clever enough to successfully define methods myself (yet). 😔

PR #730, opened 2018-09-14
Filter naively based upon hostname of the entry

### LINKED ISSUE #712: AutoFill entry list enhancements
@jhugman @sashei here are the leftover todos after #616 and #682 since that branch is merged now for TestFlight beta testing.

- [ ] filtering based on credentials provided by iOS
- [ ] Learn More links are unimplemented
- [ ] Header for the entry list need to be visible
- [ ] Doesn't work if data store / app is locked

Child of #486

jhugman:
Fixes #712.

This PR filters the entries by hostname. Some fuzzy matching is done to allow for services to use subdomains for mobile versions of the site.

This does not match on scheme.

### DIFF
--- a/CredentialProvider/Presenter/CredentialProviderPresenter.swift
+++ b/CredentialProvider/Presenter/CredentialProviderPresenter.swift
@@ -130,13 +130,8 @@ class CredentialProviderPresenter {
                         self?.view?.displayWelcome()
                     } else {
                         self?.view?.displayItemList()
-//                        guard let dismissObserver = self?.dismissObserver else { return }
-//                        self?.view?.displayAlertController(buttons: [
-//                                AlertActionButtonConfiguration(title: "OK", tapObserver: dismissObserver, style: .default)
-//                            ],
-//                                                          title: "Credential list not available yet",
-//                                                          message: "Please check back later",
-//                                                          style: .alert)
+                        let ids = serviceIdentifiers.map { $0.identifier }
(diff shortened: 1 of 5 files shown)


devinreams:
Thanks @jhugman. I'm going to ask for a simul-review from the product/UX team just to get eyes on and see if any major concerns to address here (or follow up issue, or this is perfection).

nickbrandt:
@jhugman unable to test this on BB at the moment, looks like there is a test failing

devinreams:
>Credential List does not load in apps

I have it working here using the Twitter app: biometrics prompts, unlocks, and then fills from the entry list. I'm going to update against master in case that helps (I don't think this is that far behind but maybe).

@jhugman @sashei how do you feel about the other findings? addressable today? or candidates for follow-up issues?

sashei:
@devinreams I can take a stab at the other findings today.

sashei:
@nickbrandt I think the `Show all entries` might be a stretch to do today, maybe we can have that be a follow-up item and just have the cancel button for now.

nickbrandt:
@sashei do you mean the clear for the filter/search? Basically I'm just looking for a way to get to all of their entries (even if we don't have the "show all entries" link at first). Or do you see that as a follow-up item?

devinreams:
This is not in a state to complete. We'll close this but not delete the branch and move the in progress issue back to the inbox.

PR #732, opened 2018-09-14
Hide learn more links if there is no observer for them in the item list presenter

### LINKED ISSUE #723: Hook up Learn More links in AutoFill entry list
Spun out of #712. A leftover todo after #616 and #682:

jhugman:
Fixes #723. 

Make the "Learn More" observers optionals, and then use their presence when configuring the table cells to set whether or not the "Learn More >" links should be hidden. 

Learn more observers are then set to `nil` in the `CredentialItemListPresenter`.


linuxwolf:
`ItemListPresenterSpec.swift` needs an update to treat the `learnMoreObserver` as optional.

PR #751, opened 2018-09-26
Closes #439: Scroll to top

### LINKED ISSUE #439: Reset scroll position when the sort order changes

fluffyemily:
Closes https://github.com/mozilla-lockbox/lockbox-ios/issues/439

Automatically scrolls the `tableView` in `ItemsListView` to the top after a new sort has been performed.

Sends a `ScrollAction` when the sort order is changed and then to performs the scroll to top when the `ScrollAction` is received. 

It feels like overkill, but this was the only way I could find that would ensure that the table data update had occurred before the scroll to top was performed. If the scroll to top is performed before the table data update is complete, then the top of the table changes after the scroll to top completes and so the table is not actually at the top! 

Actually, there is another way. It involved adding a very small delay (0.1secs) to the observer before the `scrollToTop` call was made. You can see this in Part 1. In all conscience I could not submit a PR where the solution was to add a random timing delay to make something work so I had to find another way. This was it. 

P.S. I know this needs tests. Still trying to figure out how to add them....

### DIFF
--- a/lockbox-ios/Presenter/ItemListPresenter.swift
+++ b/lockbox-ios/Presenter/ItemListPresenter.swift
@@ -11,6 +11,7 @@ import Shared
 
 protocol ItemListViewProtocol: AlertControllerView, SpinnerAlertView, BaseItemListViewProtocol {
     func bind(sortingButtonTitle: Driver<String>)
+    func bind(sortChanged: Driver<Setting.ItemListSort>)
     var sortingButtonEnabled: AnyObserver<Bool>? { get }
     var tableViewScrollEnabled: AnyObserver<Bool> { get }
     var pullToRefreshActive: AnyObserver<Bool>? { get }
@@ -213,14 +214,22 @@ extension ItemListPresenter {
                         return Constant.string.recent
                     }
                 }
+        
+        let itemSortScrollDriver = itemSortObservable
+                .asDriver(onErrorJustReturn: .alphabetically)
 
         view.bind(sortingButtonTitle: itemSortTextDriver)
+        view.bind(sortChanged: itemSortScrollDriver)
 
         let loginListEmptyObservable = self.dataStore.list.map { $0.isEmpty }
         let isSyncingObservable = self.dataStore.syncState.map { $0 == .Syncing }
         let enableObservable = isSyncingObservable.withLatestFrom(loginListEmptyObservable) { (isSyncing, isListEmpty) in
           return !(isSyncing && isListEmpty)
         }
+        
+//        _ = itemSortObservable.distinctUntilChanged().bind { (_itemSort) in
+//            DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + 1.0) { self.view?.scrollToTop() }
+//        }.disposed(by: self.disposeBag)
 
         enableObservable.bind(to: sortButtonObserver).disposed(by: self.disposeBag)
         enableObservable.bind(to: view.tableViewScrollEnabled).disposed(by: self.disposeBag)
--- a/lockbox-ios/Presenter/ItemListPresenter.swift
+++ b/lockbox-ios/Presenter/ItemListPresenter.swift
@@ -11,12 +11,12 @@ import Shared
 
 protocol ItemListViewProtocol: AlertControllerView, SpinnerAlertView, BaseItemListViewProtocol {
     func bind(sortingButtonTitle: Driver<String>)
-    func bind(sortChanged: Driver<Setting.ItemListSort>)
     var sortingButtonEnabled: AnyObserver<Bool>? { get }
     var tableViewScrollEnabled: AnyObserver<Bool> { get }
     var pullToRefreshActive: AnyObserver<Bool>? { get }
     var onSettingsButtonPressed: ControlEvent<Void>? { get }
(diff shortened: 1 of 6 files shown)


joeyg:
I tested this in the simulator and it worked most of the time. Occasionally it did not go all the way to the top. Here is a video: https://drive.google.com/open?id=1QkgRxXeR7PV4FF2xFbG3A54y0kK8Vm9g

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -129,6 +136,21 @@ class ItemListPresenter: BaseItemListPresenter {
...
+        
+        self.dispatcher.register
+            .filterByType(class: ScrollAction.self)
+            .subscribe({ evt in
+                guard let action = evt.element else { return }
+                switch action {
+                case .toTop:
+                    DispatchQueue.main.async {
+                        self.view?.scrollToTop()
+                    }
```
jhugman: Stick this in the `ItemListView`.

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -129,6 +136,21 @@ class ItemListPresenter: BaseItemListPresenter {
         self.setupSpinnerDisplay()
 
         let itemSortObservable = self.userDefaultStore.itemListSort
+        
+        itemSortObservable.bind(to: self.listSortedObserver).disposed(by: self.disposeBag)
```
jhugman: I'm not an Rx expect, but is it ok to re-use the same `asObservable` object here? 

```swift
self.userDefaultStore.itemListSort
```

gives you a new `Observable` each time.

fluffyemily: I am pretty sure it's OK to bind/drive the same observable more than once, but hopefully @sashei has a better idea of whether that is OK.

fluffyemily:
@joeyg can you retest to see whether this new strategy fixes the scrolling issue? If not, can you provide more details as to the scenarios in which the scrolling doesn't go all the way to the top?

fluffyemily:
@joeyg Can you give this one a go re: the sometimes failing scroll to top?

joeyg:
This works great! thanks for all your work on this @fluffyemily!

lockbox-iosTests/ItemListPresenterSpec.swift
```diff
@@ -563,8 +576,18 @@ class ItemListPresenterSpec: QuickSpec {
 
             describe("sortingButton") {
                 beforeEach {
-                    self.subject.onViewReady()
-                    self.view.fakeOnSortingButtonPressed.onNext(())
+                    let voidObservable = self.scheduler.createColdObservable([next(50, ())])
+
+                    voidObservable
+                            .bind(to: self.subject.sortingButtonObserver)
```
devinreams: @fluffyemily I tried merging `master` in to get this branch up to the latest. But I can't tell how to reconcile this test.

The changes from 4 days ago at #762 mean that `self.subject.sortingButtonObserver` doesn't quite exist here anymore and I couldn't figure this out myself.

fluffyemily: It should be just a case of keeping 

```
let itemSettingObservable: TestableObservable<Setting.ItemListSort> = self.scheduler.createColdObservable([next(50, Setting.ItemListSort.alphabetically)])
                    itemSettingObservable
                        .bind(to: self.subject.listSortedObserver)
                        .disposed(by: self.disposeBag)

                    self.scheduler.start()
```

and losing the other changes.

fluffyemily: Actually, this change causes all of my tests to fail as the fake button doesn't cause the `ScrollAction` to be fired. I may have to rethink these tests.

fluffyemily:
@devinreams I've just rebased against the upstream master and resolved the test issues. Hope this helps.

lockbox-ios/Presenter/ItemListPresenter.swift
sashei: Yes, we do that in several other places in the app.

devinreams:
>I've just rebased against the upstream master and resolved the test issues.

@fluffyemily thanks! nearly there I think.

Now I see a build error:

```
    ✗ lockbox-iosTests/ItemListViewSpec.swift:[50,22] property does not override any property from its superclass
517
    »         override var listSortedObserver: AnyObserver<Setting.ItemListSort> {
```

fluffyemily:
Damned Xcode. Why allow me to run the `ItemListPresenterSpec` test when the `ItemListViewSpec` doesn't compile? Fixed. Sorry.

devinreams:
>Fixed. Sorry.

No problem! Looks like everything is back to passing! (even though buddybuild disagrees and crashes on at least 1 test per build)

We will get this merged in shortly. Thanks for your contribution!

devinreams:
Merging despite the buddybuild failure here as tests _are_ passing, just flaking out and crashing. See #756

Issue #151, opened 2018-03-13
Copy action (UN or PW) should dismiss any active Copy overlays


changecourse:
Because our copy action is set to a standard timer, I'm able to tap between fields and have multiple states of 'copied' overlayed on top of one another, even though it's truly only the last item that I've tapped as tied to my clipboard... 

Expected behavior: 
- When I tap on one field to copy, then another, the first overlay should quickly dismiss and the new overlay should animate in as normal. They should not overlap. As for timing on the dismissal, we should pair on this and decide on the right timing @sashei

PR #761, opened 2018-10-05
151 dismiss copy overlays

michaelgill1969:
DISMISS COPY OVERLAYS

Fix issue #151 "Copy action (UN or PW) should dismiss any active Copy overlays" by adding a loop that checks for previously added StatusAlert subviews and triggers animateAlertOut if so.


joeyg:
Thanks for the fix! I’m afk but will be sure to test this out. Thanks again.

michaelgill1969:
Do I need to do something about the uispecs tests having failed?

joeyg:
@michaelgill1969 no, those are being fixed up in a different issue.

michaelgill1969:
Let me know if you would like me to adjust the timing, for example, by making the new alert wait for completion of animateAlertOut on the old alert and/or making animateAlertOut of the old alert be a factor of the fade-time constant.

sashei:
I've tagged UX for review; they should be able to address the intended behavior / timing issues.

Shared/Common/Extensions/UIViewController+.swift
```diff
@@ -42,6 +42,13 @@ protocol SpinnerAlertView {
 extension UIViewController: StatusAlertView {
     func displayTemporaryAlert(_ message: String, timeout: TimeInterval) {
         if let temporaryAlertView = Bundle.main.loadNibNamed("StatusAlert", owner: self)?.first as? StatusAlert {
+            
+            for view in self.view.subviews {
+                if view is StatusAlert {
+                    self.animateAlertOut(view, delay: timeout)
```
sashei: Is this blocking? I suspect that there may still be some overlap of alert display within the `timeout` time period. Otherwise I think this approach is fine; curious what the UX folks think of the behavior here.

sashei:
@michaelgill1969 thank you for contributing!!

devinreams:
Thank you @michaelgill1969! 🥇

PR #763, opened 2018-10-07
lock now button overrides autolock never setting

### LINKED ISSUE #693: "Never" autolock timer setting always unlocks app even after "Lock Now"
Encountered at https://github.com/mozilla-lockbox/lockbox-ios/pull/666#pullrequestreview-153960306 and briefly discussed with @sandysage and @sashei.

Placeholder to prioritize and then revisit the user story/expectations and if we want to change anything about the current behavior.

Right now (and up until now) you can set your Autolock timer setting and manually lock the app, and it will unlock. In other words, the "Never" setting means it will never be locked (via timer or otherwise).

joeyg:
Resolves #693. This could use some testing.


sashei:
ahh I see that nick has already identified one of the said edge cases..

joeyg:
@nickbrandt @sashei I made some substantial changes to this patch so it is worth a re-review. It now stores the forceLock value in `UserDefaults` so that it can be referenced when the app launches.

nickbrandt:
@joeyg this is working great for me now!

- Set auto lock to Never
- Background app, foreground app - not locked ✅ 
- Force quit app, re-open - not locked ✅ 
- Tap "Lock Now" in settings - app locked ✅ 
- Background app, foreground app - still locked ✅ 
- Force quit app, re-open - still locked ✅ 
- Unlock app, wait 5 min - still unlocked ✅

devinreams:
Small process thing: I just re-requested review from @mozilla-lockbox/mobile-engineering and dismissed @sashei's previous review since @joeyg has made "substantial" changes since the last review and is ready for another... anyone on the engineering team is free to chip in and take a look.

PR #764, opened 2018-10-07
Present biometrics on cold app open

### LINKED ISSUE #683: Display Touch ID/Face ID/Passcode prompt on cold open of app

joeyg:
Resolves #683.

This is the second attempt at this - should present biometrics on cold open but not when "Lock Now" is pressed.


Shared/Action/DataStoreAction.swift
```diff
@@ -19,6 +19,7 @@ enum DataStoreAction: Action {
     case touch(id: String)
     case add(item: LoginData)
     case remove(id: String)
+    case forceLock(locked: Bool) // Used when the "Lock Now" button is tapped
```
sashei: love the comment!

Shared/Store/BaseDataStore.swift
```diff
@@ -416,6 +427,7 @@ extension BaseDataStore {
         self.syncSubject.onNext(.ReadyToSync)
 
         // default to locked state on initialized
+        self.forceLockSubject.onNext(false)
```
sashei: If you want a default value, you could use a `BehaviorRelay`.

lockbox-ios/Presenter/WelcomePresenter.swift
```diff
@@ -131,6 +131,16 @@ extension WelcomePresenter {
...
 
         self.handleBiometrics(biometricButtonTapObservable)
+
+        // Launch the biometrics on cold start
+        let coldStartObservable = Observable.combineLatest(
+            self.accountStore.profile,
+            self.dataStore.locked.distinctUntilChanged(),
+            self.dataStore.forceLock)
+            .take(1)
+            .filter { !$0.2 }
```
sashei: nice.

PR #769, opened 2018-10-10
Update release notes

joeyg:


devinreams:
Quick note that build 2517 (not 2513) is what I just pushed to everyone on TestFlight, but its the exact same code, just from the production branch.

PR #772, opened 2018-10-11
SettingsListView: move done btn tap binding to presenter

BrianLitwin:
Connected to #209

Very similar to #762

In addition, I moved the call to set up the navbar from viewDidAppear() to viewDidLoad() so that the button was configured/available to bind before the presenter's onViewReady was called.. if thats an issue or sub-optimal, let me know. 

I didn't address the [tableView cells' selection bindings](https://github.com/mozilla-lockbox/lockbox-ios/blob/9566b2f5e7de8fc42a3df7777de0c094ea18abe3/lockbox-ios/View/SettingListView.swift#L114-L121).. didn't know if those were something you all wanted moved as well, but happy to address those too if that's the case!

### DIFF
--- a/lockbox-ios/Presenter/SettingListPresenter.swift
+++ b/lockbox-ios/Presenter/SettingListPresenter.swift
@@ -11,6 +11,7 @@ import LocalAuthentication
 protocol SettingListViewProtocol: class, AlertControllerView {
     func bind(items: Driver<[SettingSectionModel]>)
     var onLockNow: ControlEvent<Void> { get }
+    var onDoneButtonPressed: ControlEvent<Void>? { get }
 }
 
 class SettingListPresenter {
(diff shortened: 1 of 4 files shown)


joeyg:
Looks great. @BrianLitwin can you update this branch with master and we can merge it in?

BrianLitwin:
@joeyg sry but what do I need to do to make this merge-able? Was pressing the "Update branch" button what you meant?

devinreams:
>Was pressing the "Update branch" button what you meant?

@BrianLitwin Yes! We just want to make sure your code and the "production ready" code are both still compatible, tests pass, etc. Some folks like to merge from `master` using that button GitHub (or by rebasing against master, merging on command line, etc).

We'll re-run the tests and merge this in shortly. Thanks again!

BrianLitwin:
@devinreams awesome thanks!

joeyg:
Great job @BrianLitwin!

PR #783, opened 2018-10-27
Add uid to telemetry extras

### LINKED ISSUE #661: add fxa uid to extras in telemetry events
now that we have code in the desktop browser that will copy our telemetry of interest to the sync ping 
(which contains the fxa uid), we need to start including the fxa uid in our telemetry events so that we can join on that data. 

my first thought here is to add a bit to `TelemetryAction.swift` that copies the uid into the extras fileld for every event. is this the right approach?

joeyg:
Resolves #661.

### DIFF
--- a/Shared/Action/TelemetryAction.swift
+++ b/Shared/Action/TelemetryAction.swift
@@ -56,12 +56,17 @@ enum ExtraKey: String {
 }
 
 class TelemetryActionHandler: ActionHandler {
-    static let shared = TelemetryActionHandler()
-
     private let telemetry: Telemetry
+    private let accountStore: BaseAccountStore
 
-    init(telemetry: Telemetry = Telemetry.default) {
+    private var disposeBag = DisposeBag()
+
+    private var profileUid: String?
+
+    init(telemetry: Telemetry = Telemetry.default,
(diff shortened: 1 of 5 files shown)


Shared/Action/TelemetryAction.swift
```diff
@@ -57,12 +57,17 @@ enum ExtraKey: String {
...
     private let telemetry: Telemetry
+    private let accountStore: BaseAccountStore
+
+    private var disposeBag = DisposeBag()
+
+    private var profileUid: String?
 
-    init(telemetry: Telemetry = Telemetry.default) {
+    init(telemetry: Telemetry = Telemetry.default,
+         accountStore: BaseAccountStore) {
```
sashei: Can the `BaseAccountStore` have a `static shared` property that the two implementations override instead of needing to pass this in manually?

joeyg: I agree this is a bit icky but I'm not sure you can easily override static properties in Swift. It is probably worth creating an Issue around moving `BaseAccountStore` to be a protocol and cleaning up that part of the code.

sashei: Word :)

PR #784, opened 2018-10-27
add adjust event for sign in

### LINKED ISSUE #645: Adjust event tracking of "successful login"
As a marketer I want to be able to distinguish the quality of users onboarded through various acquisition channels. The quality of these users will be evaluated based on their progression through the sign-in process (sign in, confirm accounts).

In order to track this quality across acquisition channels, we will need to configure Adjust event tokens for the following unique, user specific events (tokens in parentheses):
sign-in (cuahml) 
Confirm Account (pff1na)

Adjust documentation on how to track these events can be found here: https://github.com/adjust/ios_sdk#event-tracking

joeyg:
Fixes #645.

### DIFF
--- a/lockbox-ios/Presenter/FxAPresenter.swift
+++ b/lockbox-ios/Presenter/FxAPresenter.swift
@@ -7,7 +7,6 @@ import WebKit
 import RxSwift
 import RxCocoa
 import SwiftyJSON
-import AdjustSdk
 
 protocol FxAViewProtocol: class {
     func loadRequest(_ urlRequest: URLRequest)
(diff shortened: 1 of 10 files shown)


lockbox-ios/Presenter/FxAPresenter.swift
```diff
@@ -49,23 +51,27 @@ class FxAPresenter {
 
     init(view: FxAViewProtocol,
          dispatcher: Dispatcher = .shared,
-         accountStore: AccountStore = AccountStore.shared
+         accountStore: AccountStore = AccountStore.shared,
+         adjust: Adjust
```
sashei: Similarly, is there not a good way to have an application-wide instance of `Adjust`?

PR #740, opened 2018-09-14
enable without infinite spin when no logins

sashei:
resolves #737

PR #811, opened 2018-11-14
Stop refreshing when no logins

### LINKED ISSUE #737: Attempting to enable AutoFill with a Firefox Account with no credentials spins forever
If you have an account with no synced logins (dreams+appletester) then go to Settings > Passwords > AutoFill > Enable the enable screen seems to spin infinitely.

joeyg:
Fixes #737.

Connected to #740. Bringing back this PR for testing

### DIFF
--- a/Shared/Store/CredentialProviderStore.swift
+++ b/Shared/Store/CredentialProviderStore.swift
@@ -19,6 +19,7 @@ class CredentialProviderStore {
     private let dispatcher: Dispatcher
     private let dataStore: DataStore
     private let credentialStore: CredentialIdentityStoreProtocol
+    private let accountStore: BaseAccountStore
     private let disposeBag = DisposeBag()
 
     private let _stateSubject = ReplaySubject<CredentialProviderStoreState>.create(bufferSize: 1)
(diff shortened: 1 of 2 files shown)


Shared/Store/CredentialProviderStore.swift
```diff
@@ -81,11 +84,22 @@ extension CredentialProviderStore {
     }
 
     private func refresh() {
+        Observable.combineLatest(self.dataStore.list, self.accountStore.oauthInfo, self.accountStore.profile)
```
linuxwolf: nit: this seems worthy a comment to note it's signaling there's already a list to display while waiting for the refresh ...

nickbrandt:
This worked for me (didn't encounter infinite spinner) as well after making sure to disconnect, deleting, and reinstalling clean build of 2687 as @devinreams mentioned above. Though I am still seeing entries in the dreams+appletest@mozilla.com account (entry list is populated). Once when I force killed the app and reopened it had the No Entries screen. Though now I'm consistently seeing entries...

Shared/Store/CredentialProviderStore.swift
sashei: alternately: for clarity here, I would probably do something like:
```
Observable.combineLatest(self.accountStore.oauthInfo.filterNotNull(), self.accountStore.profile.filterNotNull())
       .flatMap { self.dataStore.list }
       .filter { $0.isEmpty }
       .bind ......
```

but either way works. I did have to read the many `$0.1` type annotations a few times to understand what's going on.

devinreams:
@joeyg I'm going to update this against `master` and test again in the hopes we can get it landed for 1.3.3.

Issue #745, opened 2018-09-17
Provide an iPad specific experience


sandysage:
We continue to see user requests for a better iPad experience.

Right now 7% of our users are using the app on an iPad.

I'm opening this issue to explore the opportunities and define product requirements.

* [x] Architecture (Split View) #838
* [x] Launch screen #840 - https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c266e2d7501c0242a74cffb
* [x] Welcome Screen #840 - https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c1bd78375a3b4afa1294373
* [x] Settings #841 - https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c19385e75a3b4afa11ae404
* [x] Settings #849 #850 - 2nd level - https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c19385e2e3f282424aa66fc
* [x] Drag and Drop support #853 #854 
* [x] Onboarding Confirmation #851 - https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c266e4a839feeaf10712fa5
* [x] Webviews - https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c19385d12e8fd5096009163
* [x] Autofill UI - https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c19385e875176af8fed84a8
* [x] Empty State - https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c1aadaa0e0e0cb02dfc95b6
* [x] Autofill Onboarding #868 - https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c266e24b233e172481649ec
* [x] Lock Screen #869 - https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c1bd783a15cd3246e06806b
* [x] Autofill Setup - #868 https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c266eabb233e172481650e8
* [x] List selection state - highlight and blue #861 #862 #867 https://app.zeplin.io/project/5a4e9ff8032d0923200a2635/screen/5c19385e420c9e112aade6d7

joeyg:
@nickbrandt - One of the items in this task is webviews. Do you mind checking to see if there is anything to do there?

[...]

PR #838, opened 2018-12-30
Architecture changes for iPad Support

joeyg:
First step in providing iPad support (#745). Looking for eng feedback from @sashei @devinreams @linuxwolf. I tried a number of different approaches including:
 - Having `RootView` be an instance of `UISplitViewController`
 - Having `RootPresenter` push a `UISplitViewController` and it's nav controllers directly
 - Having `RootPresenter` push a custom `UISplitViewController` (what I settled on)

Tests need to be updated


Lockbox.xcodeproj/project.pbxproj
```diff
@@ -2134,7 +2142,7 @@
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
 				SWIFT_PRECOMPILE_BRIDGING_HEADER = NO;
 				SWIFT_VERSION = 4.2;
-				TARGETED_DEVICE_FAMILY = 1;
```
sashei: 😍

lockbox-ios/Presenter/ItemDetailPresenter.swift
```diff
@@ -96,6 +100,10 @@ class ItemDetailPresenter {
         let itemDriver = itemObservable.asDriver(onErrorJustReturn: nil)
         let viewConfigDriver = Driver.combineLatest(itemDriver, self.itemDetailStore.itemDetailDisplay)
                 .map { e -> [ItemDetailSectionModel] in
+                    if e.0 == nil {
```
sashei: would it accomplish the same thing to use `filterNil` on the `itemDriver`?

lockbox-ios/Presenter/ItemDetailPresenter.swift
```diff
@@ -105,6 +113,10 @@ class ItemDetailPresenter {
 
         let titleDriver = itemObservable
                 .map { item -> String in
+                    if item == nil {
```
sashei: ditto.

lockbox-ios/View/RootView.swift
```diff
@@ -46,16 +68,39 @@ class RootView: UIViewController, RootViewProtocol {
         self.presenter?.onViewReady()
     }
 
+    func showSidebar() {
```
sashei: This seems fine for the current design case in which we only have one view that will be displayed in split screen; I wonder if it can be generified at all for any other future cases where we'd want to have the split screen show....

lockbox-ios/View/RootView.swift
```diff
@@ -66,66 +111,93 @@ class RootView: UIViewController, RootViewProtocol {
...
             }
         case .autofillInstructions:
             if let autofillInstructionsView = UIStoryboard(name: "SetupAutofill", bundle: nil).instantiateViewController(withIdentifier: "autofillinstructions") as? AutofillInstructionsView {
-                self.currentViewController?.pushViewController(autofillInstructionsView, animated: !isRunningTest)
+                self.currentNaivgationController?.pushViewController(autofillInstructionsView, animated: !isRunningTest)
+            }
+        }
+    }
+
+    func pushSidebarView(view: MainRouteAction) {
```
sashei: These `pushXView` methods that take a `RouteAction` are getting more & more unwieldy... I wonder if it would make more sense to move away from that paradigm and have the `RoutePresenter` do the view instantiation itself as you're doing already with the `startMainStack` etc methods...

lockbox-ios/View/SplitView.swift
```diff
@@ -0,0 +1,39 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    public func showSidebar(vc: UIViewController) {
+        self.viewControllers[0] = vc
+    }
+
+    public var detailView: UINavigationController? {
+        get {
+            if self.viewControllers.count != 2 {
```
sashei: can it be `> 2` also?

lockbox-ios/View/RootView.swift
joeyg: Actually this is leftover code from a previous iteration that I wasn't happy with - it isn't called anymore. I'll just remove it :)

lockbox-ios/View/RootView.swift
joeyg: So `RootView` would expose `pushView(vc: ViewController)`, `pushSidebarView(vc: ViewController)`, `popView()` and `popToRoot()` ?

joeyg: `pushSidebarView` can probably be removed for now since we don't actually push anything there.

lockbox-ios/View/SplitView.swift
joeyg: It shouldn't https://developer.apple.com/documentation/uikit/uisplitviewcontroller/1623181-viewcontrollers

sashei: ahh cool, misremembering my `UISplitViewController` APIs...

lockbox-ios/View/RootView.swift
sashei: Yeah, that's what I was thinking. What do you think about that? Definitely open to opinions here but I think it would clean up a lot of the `RootView` code without introducing too much complexity in the `RootPresenter`.

joeyg: @sashei I updated the PR - do you mind taking a look at the new format? ViewControllers are built using a ViewFactory in `RootPresenter` and then passed to `RootView` for display.

devinreams:
@joeyg when you have a build ready to test on a real device (conflicts need to be resolved) please let me know and I’ll hammer away on my iPad Pro.

PR #841, opened 2019-01-06
iPad: Settings List

joeyg:
Depends on #838

Connected to #745

## Testing and Review Notes

- Won't have a build to test until #838 lands (with fixed tests)
- iPhone launch screen should remain unchanged
- iPad should work in portrait and landscape

## Screenshots or Videos

iPad Portrait:
simulator screen shot - ipad pro 12 9-inch - 2019-01-09 at 20 27 36

iPad Landscape:
simulator screen shot - ipad pro 12 9-inch - 2019-01-09 at 20 27 40

iPhone 8
simulator screen shot - iphone 8 - 2019-01-09 at 20 27 06


joeyg:
@nickbrandt thanks for looking into that. I'll test for the navbar soon but I expect that will be resolved with the new changes.

PR #854, opened 2019-01-16
Add drag and drop support to ItemDetailView

### LINKED ISSUE #853: iPad: add drag and drop support to username and password fields
Many apps have added “drag and drop” support for iPad apps when they’re open side by side in splitview or as an overlay. This allows users to drag the thing from App A (like a Lockbox username) over into App B (like a username field in an app or website).

https://developer.apple.com/ios/drag-and-drop/

>With systemwide drag and drop on iOS, users can quickly and easily move text, images, and files from one app to another. Simply tap and hold to pick up the content and drag it to the other app.

This would be very helpful for users on the iPad where copy/paste is not as preferable as using touch or when AutoFill doesn’t fire or work as expected.

joeyg:
Fixes #853 

Support from https://developer.apple.com/documentation/uikit/drag_and_drop/adopting_drag_and_drop_in_a_table_view

### DIFF
--- a/lockbox-ios/View/ItemDetailView.swift
+++ b/lockbox-ios/View/ItemDetailView.swift
@@ -6,6 +6,7 @@ import UIKit
 import RxSwift
 import RxCocoa
 import RxDataSources
+import CoreServices
 
 typealias ItemDetailSectionModel = AnimatableSectionModel<Int, ItemDetailCellConfiguration>
 
@@ -18,6 +19,7 @@ struct ItemDetailCellConfiguration {
     let accessibilityId: String
     let showCopyButton: Bool
     let showOpenButton: Bool
+    let dragValue: String?
 
     init(title: String,
          value: String,
@@ -26,7 +28,8 @@ struct ItemDetailCellConfiguration {
          valueFontColor: UIColor = UIColor.black,
          accessibilityId: String,
          showCopyButton: Bool = false,
-         showOpenButton: Bool = false) {
+         showOpenButton: Bool = false,
+         dragValue: String? = nil) {
         self.title = title
         self.value = value
         self.accessibilityLabel = accessibilityLabel
@@ -35,6 +38,7 @@ struct ItemDetailCellConfiguration {
         self.accessibilityId = accessibilityId
         self.showCopyButton = showCopyButton
         self.showOpenButton = showOpenButton
+        self.dragValue = dragValue
     }
 }
 
@@ -72,6 +76,7 @@ class ItemDetailView: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         self.view.backgroundColor = Constant.color.viewBackground
+        self.tableView.dragDelegate = self
         self.learnHowToEditArrow.tintColor = Constant.color.lockBoxBlue
         self.setupNavigation()
         self.setupDataSource()
@@ -153,6 +158,8 @@ extension ItemDetailView: UIGestureRecognizerDelegate {
                     cell.openButton.isHidden = !cellConfiguration.showOpenButton
                     cell.copyButton.isHidden = !cellConfiguration.showCopyButton
 
+                    cell.dragValue = cellConfiguration.dragValue
+
                     if cellConfiguration.password {
                         cell.valueLabel.font = UIFont(name: "Menlo-Regular", size: 16)
                         cell.valueLabel.preferredMaxLayoutWidth = 250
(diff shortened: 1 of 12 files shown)


lockbox-ios/View/ItemDetailView.swift
```diff
@@ -202,3 +209,20 @@ extension ItemDetailView: UIGestureRecognizerDelegate {
         }
     }
 }
+
+extension ItemDetailView: UITableViewDragDelegate {
+    func tableView(_ tableView: UITableView, itemsForBeginning session: UIDragSession, at indexPath: IndexPath) -> [UIDragItem] {
```
linuxwolf: (this is probably the wrong place to put this comment)

There are telemetry events for tracking when the username or password are copied.  We should try to record the same or similar event for drag/drop operations, if possible.

lockbox-ios/View/ItemDetailView.swift
```diff
@@ -202,3 +209,20 @@ extension ItemDetailView: UIGestureRecognizerDelegate {
...
+        let cell = tableView.cellForRow(at: indexPath)
+        let data = (cell as? ItemDetailCell)?.dragValue?.data(using: .utf8)
+        let itemProvider = NSItemProvider()
+
+        itemProvider.registerDataRepresentation(forTypeIdentifier: kUTTypePlainText as String, visibility: .all) { completion in
+            completion(data, nil)
```
linuxwolf: Can this callback also record a copy(Username/Password) telemetry event?  Or is this not the right place?

linuxwolf: Also, we really want to `touch()` the entry, to update its `timeLastUsed` and `timesUsed`.

lockbox-ios/View/ItemDetailView.swift
```diff
@@ -212,10 +212,12 @@ extension ItemDetailView: UIGestureRecognizerDelegate {
 
 extension ItemDetailView: UITableViewDragDelegate {
     func tableView(_ tableView: UITableView, itemsForBeginning session: UIDragSession, at indexPath: IndexPath) -> [UIDragItem] {
-        let cell = tableView.cellForRow(at: indexPath)
-        let data = (cell as? ItemDetailCell)?.dragValue?.data(using: .utf8)
+        let cell = tableView.cellForRow(at: indexPath) as? ItemDetailCell
+        let data = cell?.dragValue?.data(using: .utf8)
         let itemProvider = NSItemProvider()
 
+        self.presenter?.dndStarted(itemId: self.itemId, value: cell?.titleLabel.text)
```
linuxwolf: Forgive my unfamiliarity ... this will be called when the drag starts.  I take it the completion closure below doesn't work for making other calls like this.

Is there any concern about over-counting because the drop doesn't happen?  I'm guessing it's a pretty rare occurrence, so that we're fine with the "noise".

irrationalagent:
Telemetry changes are OK by me

linuxwolf:
> @joeyg this is working great, just one question:
> 
> Can we have it so when the user is pressing and holding, it doesn't initiate the copy action? Currently, if I go to hold down on a Username/Password to drag and drop, I get a copy notification.

@nickbrandt to be clear, we don't want the notification?

We definitely want to record a telemetry event and update the last used for the item; re-using the existing `CopyAction` made the most sense for those, but does carry the side-effect of that notification.

joeyg:
@nickbrandt , @linuxwolf  I updated this PR to eliminate the altering of pasteboard and showing the status alert on drag. Let me know if this works for ya.

nickbrandt:
Perfect, you're the man @joeyg !

joeyg:
> Worked like a charm on iPad Pro (Facebook and Reddit apps), and iPhone (nothing as expected), thanks @joeyg!
> 
> In my testing the drag and drop did not work on web views, but other password managers support them. Is that expected and a separate feature to file?

Thanks for testing! Do you mean you were dragging content from the webviews in Lockbox or trying to drop Lockbox credentials into a webview? Is it the same with Safari?

devinreams:
>Do you mean you were dragging content from the webviews in Lockbox or trying to drop Lockbox credentials into a webview? Is it the same with Safari?

I was dragging a username/password field from Lockbox iPad into Safari onto a web username/password form field.

Instead of Safari interpreting that as a field “fill” on drop, it instead interpreted it as a “open this content in safari” resulting in the current Safari URL literally becoming the username (in this case `firefoxlockbox@gmail.com`).

joeyg:
@devinreams updated the PR, should allow for draging into Safari now.

devinreams:
>updated the PR, should allow for draging into Safari now.

Beautiful! Thanks @joeyg 😍 

Since this is UX and code review approved: I'll merge, get multitasking updated and re-test a bit more over there #858

devinreams:
img_0026

PR #858, opened 2019-01-20
Fix multitasking mode for iPad

### LINKED ISSUE #852: iPad app does not open item details in split screen mode
Related to #745 

@joeyg you may already be aware, but wanted to add this for tracking as I didn’t see anything called out elsewhere...

If I open the iPad app as an overlay (laying on top of another open app) or in split view (eg: 40/60 split) the narrow layout shows the last view I was on (entry list, or detail).

img_0024

If the app is opened to the entry list, tapping on the entry itself does not respond. So tapping doesn’t open the entry detail.

- open an app on iPad like Safari
- open Lockbox on top of or next to Safari (overlay or split view)
- if in entry detail, tap the “Back” button to get to list view
- once in entry list, tap on any entries

Expected behavior: tapping an entry opens the detail
Actual behavior: tapping an entry shows the highlighted state (blue) but doesn’t open detail view

(That said, the search box does work on entry list. And on entry detail tapping on the table cells like username/password and buttons like copy username do work as expected)

joeyg:
Fixes #852

This fixes the multitasking mode for iPad. There are some pretty architecture changes as a result of this, namely, the list view is always a UISplitViewController now, even on iPhone. `TabletHelper` has been deprecated in favor of `SizeClassStore` but should be removed in another PR.

Connected to #745

## Testing and Review Notes

 - Should be tested on phone phone for regressions
 - Try expanding and contracting the Lockbox app while in multitasking mode
 - For some reason the navbar title has small when you expand from a stacked narrow view. This seems like a bug but probably worth fixing separately from all of this.


joeyg:
@devinreams @nickbrandt Updated the PR to fit the Lockbox logo in the narrow iPad view during multitasking. The nessie logo floats to the top in that view - but maybe that is ok for now?
[Archive.zip](https://github.com/mozilla-lockbox/lockbox-ios/files/2798576/Archive.zip)

nickbrandt:
@joeyg LGTM 👍 

img_0015

PR #887, opened 2019-03-07
Add privacy screen

### LINKED ISSUE #884: Hide the app details in application switcher
Much like we have on the Android app, we should consider masking or hiding the application view when its shown in the application switcher.

Why? For similar reasons banks and other sensitive apps do: so that private information is not disclosed by the user unintentionally when in the app switcher.

img_0044

joeyg:
Fixes #884 
simulator screen shot - iphone x - 2019-03-06 at 20 25 04


devinreams:
>Is that a constraint or something similar that can easily be added to the new PrivacyScreen layout?

If it's more than a "quick" thing then please let me know and I can merge as-is. I can file a follow-up UI issue since I think I spotted another screen in my testing doing similar...

devinreams:
I'm going to merge as-is and file a follow-up story to reconcile the storyboards per quick email with @joeyg...

PR #905, opened 2019-04-05
app start up with no network

### LINKED ISSUE #896: Support app startup with no network + no network logins access
### Acceptance Criteria
Cold starting the app without network will route appropriately to the locked screen or the item list screen. Logins synced on last use will be available.

Depends on #580

sashei:
Fixes #896 

## Testing and Review Notes

This is a _very_ baseline level of support for the app being offline. The startup is still quite slow as we wait for the FxA code to decide that the network is timing out, and could be revisited if we'd like to improve the experience per product priorities.

Possible improvements include more robust retry logic (built into pull to refresh maybe?) because we won't be able to sync again until the user hard-quits the app, having a visual indicator while waiting for the network timeouts, and including more robust network checking to route faster on startup.

## To Do

- [x] add unit tests
- [x] address iOS < 12.0
- [x] request the "UX" team perform a design review (if/when applicable)


sashei:
I'm requesting UX review here @sandysage @changecourse because I know that we aren't prioritizing our offline support yet and thus made some decisions about how great this UX is (i.e., isn't). The app now works offline, but startup is very slow (just shown a blue screen for ~20 seconds, which is the FxA timeout). I've included a few options for UX improvements above, and would like to know how y'all feel about prioritizing any of the following as part of this work:
- more robust retry logic (built into pull to refresh maybe?) because we won't be able to sync again until the user hard-quits the app [medium LOE]
- having a visual indicator while waiting for the network timeouts on startup [high LOE]
- including more robust network checking to route faster on startup [low LOE]
-

Shared/Model/SyncCredential.swift
```diff
@@ -10,6 +10,11 @@ struct SyncCredential {
     let isNew: Bool
 }
 
+let OfflineSyncCredential = SyncCredential(
+    syncInfo: SyncUnlockInfo(kid: "", fxaAccessToken: "", syncKey: "", tokenserverURL: ""),
+    isNew: false
+)
+
```
jhugman: Where are the credentials store? IIRC, the application (i.e. not the lib) is responsible for storing them, and we're doing it the key chain.

sashei: yes, we are doing that in the `AccountStore`. We don't persist the "offline" credentials in this implementation.

Shared/Store/BaseAccountStore.swift
```diff
@@ -45,6 +48,11 @@ class BaseAccountStore {
             return
         }
 
+        if !networkHelper.isConnectedToNetwork {
+            self._syncCredentials.onNext(OfflineSyncCredential)
+            return
+        }
```
jhugman: Is this ever not connected and not logged in?

Shared/Common/Helpers/NetworkHelper.swift
```diff
@@ -0,0 +1,44 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import Network
+import RxSwift
+import RxCocoa
+import SystemConfiguration
+
+class NetworkHelper {
```
jhugman: Nit: `NetworkStore` or `ConnectivityStore`, surely?

Shared/Store/BaseAccountStore.swift
sashei: no; if we don't have a stored `FirefoxAccount` we don't run this code.

Shared/Common/Helpers/NetworkHelper.swift
sashei: yawp got there locally before pushing ^.^

PR #922, opened 2019-04-23
Replace product name

joeyg:
Replaces "Firefox Lockbox" with a constant in the strings files


lockbox-ios/Presenter/SettingListPresenter.swift
```diff
@@ -173,7 +173,7 @@ extension SettingListPresenter {
                 isOn: usageDataEnabled,
                 onChanged: self.onUsageDataSettingChanged)
         let subtitle = NSMutableAttributedString(
-                string: Constant.string.settingsUsageDataSubtitle,
+            string: String(format: Constant.string.settingsUsageDataSubtitle, Constant.string.productName),
```
sashei: 👾 looks like an xcode format thing

sashei:
@joeyg would you like to fix the tests on this one? I'm happy to pick it up if you don't have time :)

joeyg:
@sashei I won’t have time today but can fix them up early next week if it is still open.

PR #932, opened 2019-04-24
fix for recurring web address launch from app

### LINKED ISSUE #889: Web Address Launching & Sync Causing Routing Problems
### Steps to reproduce
1. From entry list, go to an entry detail
2. Launch a web address in a browser (my default was Safari)
3. Use the back button in the status bar OR the app switcher to navigate back to Lockbox
4. When in Lockbox, navigate back to the entry list and perform a manual sync (or wait for an automatic sync to occur)

### Expected behavior
The Lockbox app should perform a manual sync, and that's it.

### Actual behavior
The app launches the previously selected web address (in my case, Reddit) in a browser, once the manual sync is completed

### Device and build information
- iPhone Xs
- Version 12.1.4

### Screenshare
[ScreenRecording_03-19-2019 09-35-10.MP4.zip](https://github.com/mozilla-lockbox/lockbox-ios/files/2984180/ScreenRecording_03-19-2019.09-35-10.MP4.zip)

### Notes
- I tried to replicate this issue with other entries and for whatever reason, Reddit was still the web address that was preserved and launched in browser

sashei:
Fixes #889

## To Do

- [x] add unit tests

### DIFF
--- a/lockbox-ios/Presenter/ItemDetailPresenter.swift
+++ b/lockbox-ios/Presenter/ItemDetailPresenter.swift
@@ -47,6 +47,7 @@ class ItemDetailPresenter {
             target.itemDetailStore.itemDetailId
                 .take(1)
                 .flatMap { target.dataStore.get($0) }
+                .take(1)
                 .map { item -> [Action] in
                     var actions: [Action] = []
                     if copyableFields.contains(value) {
--- a/lockbox-ios/Presenter/ItemDetailPresenter.swift
+++ b/lockbox-ios/Presenter/ItemDetailPresenter.swift
@@ -47,6 +47,8 @@ class ItemDetailPresenter {
             target.itemDetailStore.itemDetailId
                 .take(1)
                 .flatMap { target.dataStore.get($0) }
+                // The first `take(1)` call does not push a completion after it emits, so a second one is necessary here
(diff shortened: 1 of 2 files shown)


lockbox-ios/Presenter/ItemDetailPresenter.swift
```diff
@@ -47,6 +47,7 @@ class ItemDetailPresenter {
             target.itemDetailStore.itemDetailId
                 .take(1)
                 .flatMap { target.dataStore.get($0) }
+                .take(1)
                 .map { item -> [Action] in
                     var actions: [Action] = []
                     if copyableFields.contains(value) {
```
jhugman: I'm… not… sure… I… understand… how this fixes the problem. Isn't this a `RouteStore` a problem?

sashei: The subscription wasn't getting cleaned up on returning the the app, which apparently syncs on foreground (I want to clean that up in a subsequent ticket), so the `get($0)` call was pushing again. This resulted in the openwebsite action getting dispatched a second time.

jhugman: Ok, I see; thanks. In that case:

Nit: in-code comment to this effect to explain why you have two almost consecutive `take(1)`

Issue #895, opened 2019-03-28
Add DataStore tests and abstract logins support to `DataStoreSupport` class


sashei:
This will allow for better testability & serve to mirror the Android implementation better. Some work on this is being completed at #936 to aid in testing the locking functionality.

Depends on #580 

#### Acceptance Criteria
- [ ] Push the keychainwrapper code down into the DataStoreSupport
- [ ] Push the SyncUnlockInfo down into the DataStoreSupport

PR #936, opened 2019-04-25
fix backgrounding / foregrounding problems with lock

### LINKED ISSUE #898: Disable the autolock timer after disconnecting the account
### Acceptance criteria

Disconnecting an account will not result in the app being locked after backgrounding / foregrounding or cold starting after the specified amount of time.

depends on #580 
depends on #812

### LINKED ISSUE #924: Intermittent startup crash on iOS 11.4
v1.5.0(2999)
iPad Air 2 (11.4)

1. Have the app clean installed.
2. Try to open it.

Result: Intermittently, the "Welcome" screen is displayed without the "Get Started" button for a few seconds then the app crashes.

Exception Type:  EXC_CRASH (SIGKILL)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Termination Reason: Namespace SPRINGBOARD, Code 0x8badf00d
Termination Description: SPRINGBOARD, scene-create watchdog transgression: org.mozilla.ios.Lockbox exhausted real (wall clock) time allowance of 10.00 seconds |  | ProcessVisibility: Foreground | ProcessState: Running | WatchdogEvent: scene-create | WatchdogVisibility: Foreground | WatchdogCPUStatistics: ( | "Elapsed total CPU time (seconds): 5.280 (user 5.280, system 0.000), 16% CPU", | "Elapsed application CPU time (seconds): 0.576, 2% CPU" | )
Triggered by Thread:  0

### LINKED ISSUE #926: Able to unlock the app without being signed in
v1.5.0(2999)
iPad Air 2 (11.4)

1. Have the biometrics enabled.
2. Install the app and open it.
3. While in the "Welcome" screen, background then foreground the app.
         => The user is prompt for biometrics.
4. Enter the biometrics without signing in previously.

Result: The app is unlocked. 

depends on #898

sashei:
Fixes #898
Fixes #926 
Fixes #924 

Connected to #895

## Testing and Review Notes
This also fixes a bug introduced at #916 in which using the Lock Now button and then foregrounding before the autolock time would unlock the app.

For codereviewers: I got a head start on #895 in this issue so that all of the locking behavior can be more explicitly tested.

## To Do

- [x] add “WIP” to the PR title if pushing up but not complete nor ready for review
- [x] add unit tests

### DIFF
--- a/Shared/Common/Helpers/DataStoreSupport.swift
+++ b/Shared/Common/Helpers/DataStoreSupport.swift
@@ -17,6 +17,7 @@ public protocol LoginsStorageProtocol {
     func list() throws -> [LoginRecord]
 }
 
+// We decorate the LoginsStorage with the LoginsStorageProtocol so that it's easy to mock and inject for unit testing.
 extension LoginsStorage: LoginsStorageProtocol { }
 
 class DataStoreSupport {
--- a/Shared/Common/Helpers/DataStoreSupport.swift
+++ b/Shared/Common/Helpers/DataStoreSupport.swift
@@ -1,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import MozillaAppServices
+
+public protocol LoginsStorageProtocol {
+    func close()
+    func isLocked() -> Bool
+    func ensureUnlocked(withEncryptionKey key: String) throws
+    func ensureLocked()
+    func sync(unlockInfo: SyncUnlockInfo) throws
+    func wipeLocal() throws
+    func get(id: String) throws -> LoginRecord?
+    func touch(id: String) throws
+    func list() throws -> [LoginRecord]
+}
+
+extension LoginsStorage: LoginsStorageProtocol { }
+
+class DataStoreSupport {
+    static let shared: DataStoreSupport = DataStoreSupport()
+
+    open func createLoginsStorage(databasePath: String) -> LoginsStorageProtocol {
+        return LoginsStorage(databasePath: databasePath)
+    }
+}
--- a/Shared/Store/BaseDataStore.swift
+++ b/Shared/Store/BaseDataStore.swift
@@ -107,23 +107,22 @@ class BaseDataStore {
     }
 
     // From: https://github.com/mozilla-lockbox/lockbox-ios-fxa-sync/blob/120bcb10967ea0f2015fc47bbf8293db57043568/Providers/Profile.swift#L168
-    internal static var loginsKey: String? {
+    internal var loginsKey: String? {
         let key = "sqlcipher.key.logins.db"
-        let keychain = KeychainWrapper.sharedAppContainerKeychain
-        if keychain.hasValue(forKey: key) {
-            return keychain.string(forKey: key)
+        if self.keychainWrapper.hasValue(forKey: key) {
+            return self.keychainWrapper.string(forKey: key)
         }
 
         let Length: UInt = 256
         let secret = Bytes.generateRandomBytes(Length).base64EncodedString(options: [])
-        keychain.set(secret, forKey: key, withAccessibility: .afterFirstUnlock)
+        self.keychainWrapper.set(secret, forKey: key, withAccessibility: .afterFirstUnlock)
         return secret
     }
 
     var unlockInfo: SyncUnlockInfo?
 
     init(dispatcher: Dispatcher = Dispatcher.shared,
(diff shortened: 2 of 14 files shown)


Shared/Store/BaseDataStore.swift
```diff
@@ -269,19 +262,44 @@ extension BaseDataStore {
 
 // locking management
 extension BaseDataStore {
-    public func unlock() {
+    private func setupAutolock() {
+        let emitsOnNewLifecycleObservable = Observable.combineLatest(self.lifecycleStore.lifecycleEvents, self.storageState)
+                .distinctUntilChanged { (lhLifecycleState, rhLifecycleState) -> Bool in
+                    return lhLifecycleState.0 == rhLifecycleState.0
+                }
```
jhugman: Squinting, I wander if this could be better done with a 

```swift
    .distinctUntilChanged()
    .switchMap {}
```

but I'm not completely sure.

Shared/Store/LifecycleStore.swift
```diff
@@ -13,7 +13,7 @@ class LifecycleStore {
     private let dispatcher: Dispatcher
     private let _lifecycleFilter = PublishSubject<LifecycleAction>()
 
-    public var lifecycleFilter: Observable<LifecycleAction> {
+    public var lifecycleEvents: Observable<LifecycleAction> {
```
jhugman: 😻

Shared/Store/BaseDataStore.swift
```diff
@@ -157,26 +157,19 @@ class BaseDataStore {
...
-                        self.handleLock()
-                    case .upgrade(let previous, _):
-                        if previous <= 2 {
-                            self.handleLock()
-                        }
-                    case .shutdown:
-                        self.shutdown()
-                    default:
-                        break
-                    }
```
jhugman: I understand replacing the `filterByType` with the `lifeCycleStore.lifecycleEvents`.

I don't understand how/why you can get rid of _all_ of the `switch` statement. Especially the `upgrade` actions.

Shared/Common/Helpers/DataStoreSupport.swift
```diff
@@ -0,0 +1,28 @@
...
+}
+
+extension LoginsStorage: LoginsStorageProtocol { }
+
+class DataStoreSupport {
+    static let shared: DataStoreSupport = DataStoreSupport()
+
+    open func createLoginsStorage(databasePath: String) -> LoginsStorageProtocol {
+        return LoginsStorage(databasePath: databasePath)
+    }
```
jhugman: I think you just made a factory :)

Shared/Common/Helpers/DataStoreSupport.swift
```diff
@@ -0,0 +1,28 @@
...
+    func ensureUnlocked(withEncryptionKey key: String) throws
+    func ensureLocked()
+    func sync(unlockInfo: SyncUnlockInfo) throws
+    func wipeLocal() throws
+    func get(id: String) throws -> LoginRecord?
+    func touch(id: String) throws
+    func list() throws -> [LoginRecord]
+}
+
+extension LoginsStorage: LoginsStorageProtocol { }
```
jhugman: IIUC, you're decorating a `MozillaAppServices` class with our own `LoginsStorageProtocol`.

If this is the case, could you add a comment to that effect; there's a point of confusion about the empty extension block.

Shared/Store/BaseDataStore.swift
sashei: No, using `combineLatest` means that the same version of the lifecycle will get pushed thru with each subsequent state, which isn't desirable. `zip` isn't quite right either; we want the most updated state variable when getting a new lifecycle action.

Shared/Common/Helpers/DataStoreSupport.swift
sashei: 🤐

Shared/Store/BaseDataStore.swift
sashei: A lot of them got moved into the `setupAutoLock` method, and then it seemed like overkill to have a switch statement for two cases so I split those out into their own subscriptions, although I think that's more of a personal preference than a design requirement.

re: removing the `upgrade` code, I'm OK with cleaning up that dead code as >99% of our active users have upgraded to a higher version already, and the workaround for missing the upgrade code is that they will have to log in again.

Shared/Store/BaseDataStore.swift
sashei: oh wait, I see what you're saying

Shared/Store/BaseDataStore.swift
jhugman: :+1:

PR #943, opened 2019-04-30
Remove product name from storyboards

joeyg:
We had instances where Firefox Lockbox was hardcoded in storyboard. This PR removes them.


lockbox-ios/Common/Resources/Constants.swift
```diff
@@ -98,4 +98,9 @@ extension Constant.string {
     static let reauthenticationRequired = NSLocalizedString("reauth_required", value: "Reauthentication Required", comment: "Title of dialog box displayed when users need to reauthenticate")
     static let appUpdateDisclaimer = NSLocalizedString("app_update_explanation", value: "Due to a recent app update, we will need you to sign in again. Apologies for the inconvenience.", comment: "Message in dialog box when users need to reauthenticate explaining application update")
     static let continueText = NSLocalizedString("continue", value: "Continue", comment: "Button title when agreeing to proceed to access Lockbox.")
+    static let accessProduct = NSLocalizedString("welcome.accessProduct", value: "To use %@, you’ll need a Firefox Account with saved logins.", comment: "Access message displayed to user on welcome screen")
```
sashei: Do we need to parameterize things like `Firefox Account` in a similar way @Delphine ?

lockbox-ios/View/WelcomeView.swift
```diff
@@ -29,6 +29,7 @@ class WelcomeView: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         self.roundButtons()
+        self.setupStrings()
```
sashei: 🛁

lockbox-ios/Common/Resources/Constants.swift
Delphine: We've left Firefox Accounts as localizable for other projects, so we should be good for that :)

PR #959, opened 2019-05-08
Fix some strings being duped in export

joeyg:
Fixes some strings being duped when exporting https://github.com/mozilla-l10n/lockwiseios-l10n/issues/4#issuecomment-488257821


joeyg:
Did not work :(

PR #961, opened 2019-05-08
fix autofill issues with rust logins

### LINKED ISSUE #950: Lockbox 1.5.0 only shows a blue screen
Ever since the update to 1.5.0 I can't get Lockbox to work. The first update was crashing, but the last couple of versions just resulted in a blue screen after startup. Currently (build 3135) cold opening the app from the home screen goes from the splash screen to a blue page, without an unlock prompt or anything else. Starting from an autofill flow I see the correct header (cancel button, app name and search box), but an empty white space where the account list would appear.

### LINKED ISSUE #954: Autofill only fills sometimes on 1.5.0
Seems I'm encountering two things with 1.5.0 now:

1. after first setup and run with the app and system settings, the app fills nicely. but after backgrounding and letting it sit a while (maybe background crash?) then the autofill keyboard bar does show matching credentials, but tapping it _does not fill_ in the fields

2. after tapping the autofill entry in the keyboard bar once or a few times, autofill seems to crash or lose its data entirely and I don't see autofill matches on the keyboard bar again

sashei:
Fixes #954

## Testing and Review Notes

I tested this primarily by following the steps to reproduce outlined by @devinreams in the bug report.

## To Do

- [x] add “WIP” to the PR title if pushing up but not complete nor ready for review
- [x] add unit tests

### DIFF
--- a/Shared/Store/BaseDataStore.swift
+++ b/Shared/Store/BaseDataStore.swift
@@ -93,7 +93,7 @@ class BaseDataStore {
     }
 
     public var locked: Observable<Bool> {
-        return self.storageState.map { $0 != LoginStoreState.Unlocked }
+        return self.storageState.map { $0 == LoginStoreState.Locked }
     }
 
     public var storageState: Observable<LoginStoreState> {
@@ -260,7 +260,6 @@ extension BaseDataStore {
 extension BaseDataStore {
     private func setupAutolock() {
         self.lifecycleStore.lifecycleEvents
-                .distinctUntilChanged()
                 .filter { $0 == .background }
                 .flatMap { _ in self.storageState }
                 .take(1)
@@ -272,7 +271,6 @@ extension BaseDataStore {
                 .disposed(by: disposeBag)
 
         self.lifecycleStore.lifecycleEvents
-                .distinctUntilChanged()
                 .filter { $0 == .foreground }
                 .flatMap { _ in self.storageState }
                 .take(1)
@@ -359,7 +357,6 @@ extension BaseDataStore {
 
     private func shutdown() {
         self.loginsStorage?.close()
-        self.loginsStorage = nil
     }
 
     private func initializeLoginsStorage() {
--- a/Shared/Store/BaseDataStore.swift
+++ b/Shared/Store/BaseDataStore.swift
@@ -161,8 +161,9 @@ class BaseDataStore {
         self.lifecycleStore.lifecycleEvents
                 .filter { $0 == .foreground }
                 .subscribe(onNext: { [weak self] _ in
-                    if let _ = self?.loginsStorage {
+                    guard let _ = self?.loginsStorage else {
                         self?.initializeLoginsStorage()
+                        return
                     }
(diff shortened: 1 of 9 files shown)


devinreams:
@sashei in the OG description this says fixed 950, do you mean #954?

sashei:
@devinreams yes!!

sashei:
@devinreams did you let the device sit with the credential provider open explicitly (i.e., the entries list from selecting Lockbox on the keyboard?) Or just left it after using the quick-type bar to fill?

devinreams:
>did you let the device sit with the credential provider open explicitly

No, the app was fully backgrounded and the credential provider was not open/active.

Shared/Store/BaseDataStore.swift
```diff
@@ -161,7 +161,9 @@ class BaseDataStore {
         self.lifecycleStore.lifecycleEvents
                 .filter { $0 == .foreground }
                 .subscribe(onNext: { [weak self] _ in
-                    self?.initializeLoginsStorage()
+                    if (self?.loginsStorage == nil) {
+                        self?.initializeLoginsStorage()
+                    }
```
jhugman: Nit: avoid typing `nil`.

```swift
guard let _ = self?.loginsStorage else {
   self?.initializeLoginsStorage()
}
```

Shared/Store/BaseDataStore.swift
```diff
@@ -186,12 +188,12 @@ class BaseDataStore {
...
-            .map { items -> LoginRecord? in
-                return items.filter { item in
-                    return item.id == id
-                    }.first
+    public func get(_ id: String) -> LoginRecord? {
+        do {
+            return try self.loginsStorage?.get(id: id)
+        } catch let error {
+            print("Sync15: \(error)")
+            return nil
```
jhugman: AFAICT, this is the crux of the matter. 

It feels quite synchronous _-y_ and MVC _-y_, which makes me nervous. Are there any restrictions you'd like to make on how this method should be called? i.e. if this represents an exception to the rules around how we manage state, how can we make the exception smaller? e.g. perhaps if we expose most of the app to the previous way of getting state, but only use this new synchronous route for _that one time_ when we have too. 

Also: could you comment that this is unusual, and why we're doing it in not the usual way.

sashei: :-/ this is a great point, and actually doesn't resolve the bug in the first place anyway. I realized this `get` method was there and doing the sql lookup under the hood was almost definitely more performant than what we had before so I was just optimizing. It _should_ be async though, so I'm changing it.

jhugman:
Changing back to the async `get(id)` has made the patch a lot smaller. Good job!

devinreams:
I've been testing reddit.com/login off and on for the past 24 hours and autofill has always been successfully showing keyboard matches and filling on first attempt. I just tried once more again this morning and it worked A-OK still.

So I haven't reproduced the issue since yesterday toggling off/on AutoFill in system settings.

PR #968, opened 2019-05-13
Installation and local build need to mention rust

Pike:
Rust needs to have the iOS targets installed.

Grabbed the list from https://mozilla.github.io/firefox-browser-architecture/experiments/2017-09-06-rust-on-ios.html

Also `brew install swift-protobuf`


joeyg:
Is lockwise's swift version not matching the version building the application-services dependency? Previously we have worked to avoid having developers install rust in order to build lockbox.

Pike:
```
*** Downloading application-services.framework binary at "0.27.0.1 Xcode 10.2 build of 0.27"
***  Skipped installing application-services.framework binary due to the error:
	"Incompatible Swift version - framework was built with 5.0.1 (swiftlang-1001.0.82.4 clang-1001.0.46.5) and the local version is 5.0 (swiftlang-1001.0.69.5 clang-1001.0.46.3)."
```

That might also explain why the XCode build doesn't work afterwards, it doesn't find `MozillaAppServices`. Which are in `Carthage/Build/iOS/Static/MozillaAppServices.framework/` locally, I think. I wonder if the `Static` is throwing it off?

joeyg:
@Pike do you have Xcode 10.2.1? That is the version that ships Swift 5.0.1

sashei:
@Pike yeah, we pin developers to a specific Xcode version rather than enforcing our contributors get their rust env setup

Pike:
So, with some luck, I got to a working local build. There's also been a few github API failures, which can randomly also hit the rust parts. Killing that and trying again helped.

sashei:
@Pike 😳 thanks for updating our version documentation!!

devinreams:
TIL 🌈 

I feel like I got lucky and actually "received" this Xcode update, otherwise would have been in the same spot. Sorry and thank you @Pike

PR #1003, opened 2019-05-22
fix issue where locking only happens once per app lifecycle

### LINKED ISSUE #1002: Autolock timer does not work after STR
Build: 1.6.0(3339)
Device: iPhone Xr
iOS: 12.2

Preconditions:
- Biometrics to unlock enabled

Steps to reproduce:
1. Launch Lockwise
2. Unlock Lockwise
3. Lock the device while app is in foreground (wait the amount of time it's need in order for the app to lock)
4. Unlock the device -> The app will be unlocked
5. Backround the app and lock the device
6. Wait the amount of time it's need in order for the app to lock
7. Unlock the device
8. Open the app

Actual results:
- The app will be unlocked.

sashei:
Fixes #1002 

## Testing and Review Notes

The bug description is very useful! This can also be reproduced by simply backgrounding, waiting for the app to lock once, and then trying to get it to lock in the background (NOT hard quitting) a second time.

## To Do

- [x] add unit tests
- [ ] make sure CI builds are passing (e.g.: fix lint and other errors)

### DIFF
--- a/Shared/Store/BaseDataStore.swift
+++ b/Shared/Store/BaseDataStore.swift
@@ -271,22 +271,25 @@ extension BaseDataStore {
 extension BaseDataStore {
     private func setupAutolock() {
         self.lifecycleStore.lifecycleEvents
-                .filter { $0 == .background }
-                .flatMap { _ in self.storageState }
-                .take(1)
-                .subscribe(onNext: { [weak self] state in
-                    guard state == .Unlocked else { return }
-
-                    self?.autoLockSupport.storeNextAutolockTime()
-                })
-                .disposed(by: disposeBag)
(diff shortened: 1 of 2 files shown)


Shared/Store/BaseDataStore.swift
```diff
@@ -271,22 +271,25 @@ extension BaseDataStore {
 extension BaseDataStore {
     private func setupAutolock() {
         self.lifecycleStore.lifecycleEvents
-                .filter { $0 == .background }
-                .flatMap { _ in self.storageState }
-                .take(1)
```
linuxwolf: TAAAKKE!

PR #1031, opened 2019-05-29
make sure credential provider works appropriately when not authenticated

### LINKED ISSUE #953: Lockbox incorrectly assumes its not logged in, hijacks screen and breaks original app that launched it
When lockbox is not authed,  immediately on launch it displays the pop up message saying it must be logged in first. If the "ok" button isn't pressed, it eventually does the Face ID recognition and logs in, but the popup is still there. If you press "ok", it closes lockbox but goes back to an unusable version of the original app that opened it.

I'm able to recreate it quite easily and this happens constantly. It stopped happening for a while and came back again recently.

I have a video I can share privately of the entire workflow in producing the bug.

### LINKED ISSUE #1005: Lockwise autofill can be prompted after disconnecting account in Lockwise
Build: 1.6.0(3339)
Device: iPhone 7
iOS: 12.0

Preconditions:
- Set up autofill password for Lockwise and keychain in iOS settings
- Disconnect account in Lockwise settings

Steps to reproduce:
1. Launch Lockwise in Welcome Screen
2. Tap "Get started"
3. Manually type the email address 
4. When prompted to type the password, try to auto fill the password via Lockwise.

Actual results:
- Lockwise autofill is unresponsive.

sashei:
Fixes #953 
Fixes #1005 

## Testing and Review Notes

The bug description in 953 is useful & descriptive. This should also resolve the "lockup" issue experienced in #1005.

## To Do

- [x] add unit tests

### DIFF
--- a/CredentialProvider/Presenter/CredentialProviderPresenter.swift
+++ b/CredentialProvider/Presenter/CredentialProviderPresenter.swift
@@ -48,13 +48,14 @@ class CredentialProviderPresenter {
         self.credentialProviderStore = credentialProviderStore
 
         self.accountStore.syncCredentials
-            .bind { [weak self] (syncInfo) in
+            .map { syncInfo -> Action in
                 if let credentials = syncInfo {
-                    self?.dispatcher.dispatch(action: DataStoreAction.updateCredentials(syncInfo: credentials))
+                    return DataStoreAction.updateCredentials(syncInfo: credentials)
                 } else {
-                    self?.dispatcher.dispatch(action: DataStoreAction.reset)
+                    return DataStoreAction.reset
                 }
             }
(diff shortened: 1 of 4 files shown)


duoi:
Are there situations where tokens expire and the login to Lockbox needs to be re-authed?

How is token expiry determined? i.e, if an attempt to verify the token is unsuccessful (server down or client can't connect) then is a flag set somewhere on the app to ask it to reauth on open?

What I'm trying to get at is what state does the app need to be in where the pop-up would be displayed to re-auth a user? And is that state being checked when the app is opened via autofill? Why/why not?

My takeaway from this, especially as to how the prompt comes up as soon as the app is opened, that a flag is being set somewhere incorrectly that the rest of the app doesn't seem to give priority to.

Issue #1009, opened 2019-05-23
Several strings are not localized in onboarding and autofill screens


TheoChevalier:
Multiple strings are either hardcoded,  haven’t been exposed to the xliff file, or are not used. Adding screenshots below

- [x] “Deselect iCloud keychain” → Not exposed to Pontoon (#1032)
- [x] “Firefox Lockbox will use Face ID to sign you in to the app” → Not exposed to Pontoon (#992)
- [ ] “Sync between devices with secure 256-bits encryption” → Exposed, but not used
- The headline in the instruction screen could use a line break, too (moved to #1027 #1029 )

![IMG_5183](https://user-images.githubusercontent.com/1294206/58275991-9c28d980-7d96-11e9-8b04-c1ecefab8e1e.PNG)

![IMG_5189](https://user-images.githubusercontent.com/1294206/58276987-080c4180-7d99-11e9-880a-0391c71fbda5.PNG)

![IMG_5190](https://user-images.githubusercontent.com/1294206/58276869-ce3b3b00-7d98-11e9-89cf-0bfca38e530a.PNG)

TheoChevalier:
Morphing this bug to include additional missing strings:

![IMG_5189](https://user-images.githubusercontent.com/1294206/58276987-080c4180-7d99-11e9-880a-0391c71fbda5.PNG)

![IMG_5190](https://user-images.githubusercontent.com/1294206/58276869-ce3b3b00-7d98-11e9-89cf-0bfca38e530a.PNG)

devinreams:
Thanks for the report @TheoChevalier. We'll look at this shortly...

devinreams:
>“Sync between devices with secure 256-bits encryption” → Exposed, but not used

@sashei I can't tell why this one may not be connecting up:

>“Sync between devices with secure 256-bits encryption” → Exposed, but not used

Anything obvious jump out at you? I'll look more tomorrow

sashei:
@devinreams we are doing manual interpolation to put the link in for `256-bit encryption`. Looking at some languages, we expect the storyboard text to match exactly a _different_ piece of text that's localized as `Constant.string.onboardingSecurityPostfix`. The string formatting is definitely failing because the strings (the one in the storyboard + the one in the generic `Localizable.strings`) don't line up. I think we need to tweak how we are getting that string interpolated + localized to work (although I'm not sure why it would fall back to English when the string formatting fails??).

[...]

PR #992, opened 2019-05-17
Remove product name from strings

joeyg:
Fixes #991

PR #1032, opened 2019-05-29
explictly add "Deselect iCloud Keychain" strings to all locales

devinreams:
Connected to #1009


devinreams:
>Cool, will we be able to get these back from Pontoon as well?

That's the idea! I'll just run the export script after this and the other #992 fix merges on master.

>I think iCloud Keychain is called different things on different platforms.

Yeah and that's noted in the hint to localizers comment.

devinreams:
Hmm I just ran the export script and it didn't pull this out yet. 🤔

devinreams:
OK that export worked now:
https://github.com/mozilla-l10n/lockwiseios-l10n/compare/master...devinreams:20190529_1637?expand=1#diff-b1b0d455e51cc8377d5f52b6ad5f164eR731

I forgot I made more changes upstream to make that one a UILabel while doing wrapping fixes. 🚀

PR #1045, opened 2019-06-04
implement sync timeout

### LINKED ISSUE #1025: Sync timeout
Also do #726

- [x] investigate where/why we wouldn't get a timeout error from a-s
- [x] introduce a timeout in sync (so no infinite spinner)
- [x] catch this condition and log it to sentry
- [x] show an empty state with a message "Syncing has timed out."

sashei:
Fixes #1025 

## Testing and Review Notes
Not sure the best way to test this other than Network Link Conditioner / being on a very slow connect.

## To Do

- [ ] make sure CI builds are passing (e.g.: fix lint and other errors)

### DIFF
--- a/Shared/Store/BaseDataStore.swift
+++ b/Shared/Store/BaseDataStore.swift
@@ -236,6 +236,11 @@ extension BaseDataStore {
         }
 
         queue.async {
+            self.queue.asyncAfter(deadline: .now() + 20, execute: {
+                // this block serves to "cancel" the sync if the operation is running slowly
+                self.syncSubject.onNext(.Synced)
+            })
+
             do {
                 try self.loginsStorage?.sync(unlockInfo: syncInfo)
             } catch let error as LoginStoreError {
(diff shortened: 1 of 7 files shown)


Shared/Store/BaseDataStore.swift
```diff
@@ -229,21 +231,29 @@ extension BaseDataStore {
...
-            self.syncSubject.onNext(SyncState.Syncing)
+            self.syncSubject.accept(SyncState.Syncing)
         } else {
-            self.syncSubject.onNext(SyncState.Synced)
+            self.syncSubject.accept(SyncState.Synced)
             return
         }
 
         queue.async {
+            self.queue.asyncAfter(deadline: .now() + 20, execute: {
```
linuxwolf: this queues for 20 seconds from `.now()`, correct?  I ask because I am not super-familiar with the interface, and most other systems seem to use milliseconds.

Also, is it worth making `20` a constant?

sashei: yes, it queues for 20 seconds from now. Constant is a good idea!

PR #1057, opened 2019-06-14
Display the translated string for text view

### LINKED ISSUE #1009: Several strings are not localized in onboarding and autofill screens

### LINKED ISSUE #1061: Change the secure text interpolation on the `OnboardingConfirmation` screen
Currently, we are assuming that the localized version of `256-bit encryption` in our string constants will always line up with the localized version of the Storyboard string. Rather than using _any_ strings from the `OnboardingConfirmation.storyboard`, this ticket will see creating a new strings with base values of `Sync between devices with secure %@` and `256-bit encryption`, allowing us to safely a) interpolate the link text and b) know that the link text will have a `range` in the destination string to be "linked" to.

joeyg:
Fixes #1061 

The middle piece of text on this onboarding screen is handled differently due to the highlighting of "256-bit encryption" in the string. Previously the code was pulling the English text from the label, highlighting it and then assigning it. That resulted in no translations being shown.


joeyg:
> I tested this PR in German and Spanish; it does not successfully do the string interpolation and the string is still displayed in English. I've included more details in the new ticket #1061 and pointed this PR there; let me know if you have questions or suggestions about my proposed solution :)

@sashei I don't think this is testable in that way until the strings have gone through the translation. Once I moved the string from the Storyboard until the Constant's file it will get a new translation id.

PR #1058, opened 2019-06-16
Fix navbar on edit screen

joeyg:
This is a PR against the edit-layout screen branch (Details [here](https://github.com/mozilla-lockwise/lockwise-ios/pull/981#issuecomment-502480430)) It fixes two issues with the navbar:

- Changing the text from "Edit" to "Save" resulted in the text not fitting in the UIBarButtonItem. `sizeToFit()` needed to be called to resize the button.
- The "Cancel" button still have the back arrow "<"


lockbox-ios/View/ItemDetailView.swift
```diff
@@ -79,15 +79,25 @@ extension ItemDetailView: ItemDetailViewProtocol {
...
     }
 
     var leftButtonText: AnyObserver<String?> {
         // swiftlint:disable:next force_cast
         return (self.navigationItem.leftBarButtonItem!.customView as! UIButton).rx.title().asObserver()
     }
 
+    var leftButtonIcon: AnyObserver<UIImage?> {
+        // swiftlint:disable:next force_cast
+        return (self.navigationItem.leftBarButtonItem!.customView as! UIButton).rx.image().asObserver()
```
devinreams: @joeyg I'm not sure how but I think the left button is still appearing during selected/highlighted states.

Screen Shot 2019-06-28 at 2 46 39 PM


In other words, if you tap the "Cancel" button and hold it down you'll see the back arrow "<" appear.

I couldn't figure out how I could fix this myself though...

devinreams: Filed follow up at #1082

PR #1059, opened 2019-06-16
Confirm dialog for edit screen

joeyg:
This is a PR against the edit-layout screen branch (Details here). It adds the [discard changes dialog](https://mozilla.invisionapp.com/share/BQEUSO0S2#/screens/348017208).


lockbox-ios/Presenter/ItemDetailPresenter.swift
```diff
@@ -151,9 +158,30 @@ class ItemDetailPresenter {
 
     private func setupNavigation() {
         self.view?.leftBarButtonTapped
-                .withLatestFrom(self.itemDetailStore.isEditing) { _, editing -> Action in
-                    return editing ? ItemDetailDisplayAction.viewMode : MainRouteAction.list
+                .withLatestFrom(self.itemDetailStore.isEditing) { _, editing -> Action? in
+                    if editing {
+                        self.view?.displayAlertController(
```
sashei: I think we want to have more logic around whether there are changes to the fields in the editor view rather than blindly throwing up the dialog when leaving the editor view. @changecourse can you confirm?

lockbox-ios/Presenter/ItemDetailPresenter.swift
```diff
@@ -151,9 +158,30 @@ class ItemDetailPresenter {
...
+                                    style: .destructive)
+                            ],
+                            title: Constant.string.discardChangesTitle,
+                            message: Constant.string.discardChangesMessage,
+                            style: .alert,
+                            barButtonItem: nil)
+                    } else {
+                        return MainRouteAction.list
+                    }
+                    return nil
```
sashei: why is there a `return nil` case? the `if editing` case will never fall through... is this a weird swift thing?

lockbox-ios/Presenter/ItemDetailPresenter.swift
joeyg: yeah that's definitely a unclear point when going off invision screens. If we need a "did anything change" check here that's fine but not sure what UX was intended.

devinreams: I think you are both right.

But for sake of getting to complete, I'm going to move this open question to confirm/do as a separate issue: #1071

lockbox-ios/Presenter/ItemDetailPresenter.swift
devinreams: Without it we see the error:

>Missing return in a closure expected to return 'Action?'

devinreams:
Works as expected, and code looks OK here. Thank you @joeyg! 🙇 

I'll go ahead and merge this into the bigger layout PR #1060 so we can review the 'edit mode' a bit more holistically. 🌈

PR #1090, opened 2019-07-27
Sync telemetry

joeyg:
iOS version of https://github.com/mozilla-lockwise/lockwise-android/pull/835

### DIFF
--- a/Shared/Action/DataStoreAction.swift
+++ b/Shared/Action/DataStoreAction.swift
@@ -16,6 +16,59 @@ enum DataStoreAction: Action {
     case sync
     case touch(id: String)
     case delete(id: String)
+    case syncEnd
+    case syncTimeout(error: String)
+    case syncError(error: String)
+}
+
+extension DataStoreAction: TelemetryAction {
+    var eventMethod: TelemetryEventMethod {
+        switch(self) {
+        case .updateCredentials:
+            return .update_credentials
+        case .lock:
+            return .lock
+        case .unlock:
+            return .unlock
+        case .reset:
+            return .reset
+        case .sync:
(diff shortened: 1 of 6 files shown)


irrationalagent:
@eliserichards [here is some docs](https://github.com/mozilla-lockwise/lockwise-ios/blob/master/docs/metrics.md) I'll review this today as well.

Shared/Action/DataStoreAction.swift
```diff
@@ -16,6 +16,59 @@ enum DataStoreAction: Action {
...
+        case .updateCredentials:
+            return .update_credentials
+        case .lock:
+            return .lock
+        case .unlock:
+            return .unlock
+        case .reset:
+            return .reset
+        case .sync:
+            return .sync
```
irrationalagent: given that there is also sync timeout end and error, should this be sync start or similar?

Shared/Action/DataStoreAction.swift
```diff
@@ -16,6 +16,59 @@ enum DataStoreAction: Action {
...
+        return .datastore
+    }
+
+    var value: String? {
+        switch(self) {
+        case .delete(let id):
+            return id
+        case .touch(let id):
+            return "ID: \(id)"
+        case .syncError(let error):
```
irrationalagent: according the telemetry library there is a max string length of 50 https://github.com/mozilla-mobile/telemetry-ios/blob/aac1e1873134ea69f992e3bbc45b431d6564d64c/Telemetry/TelemetryEvent.swift#L8 its unclear to me what would happen if we pass it something longer,e.g.  if an error message is really long

PR #1106, opened 2019-10-31
Ios 13 ui

kaylagalway:
Fixes [1099](https://github.com/mozilla-lockwise/lockwise-ios/issues/1099)
Fixes [1097](https://github.com/mozilla-lockwise/lockwise-ios/issues/1097)
Fixes [1113](https://github.com/mozilla-lockwise/lockwise-ios/issues/1113)

- Update codebase to newest version of Swift(5.1.2) and Xcode(11.2.1) and get everything functioning on iOS 13
- Fix navigation bar color and size issues caused by changes in iOS 13 on both iPhone and iPad
- Fix issue where logins do not appear on first launch due to iOS 13 change where `traitCollectionDidChange` is no longer called on initial view layout
- Fix all UI tests failing due to iOS 13 Safari updates

## Testing and Review Notes

- Launch app on both iOS and iPad
- Verify that header is rendering correctly on the Item List page and Item Detail page (screenshots below). Navigation bar background color should be purple and navigation title text should be small on Item List page and large on Item Detail page.  
- Verify login credential table view cells display on initial launch on all devices

## Screenshots or Videos

ItemListView
ItemDetailView


LockwiseXCUITests/LockwiseScreenGraph.swift
```diff
@@ -268,7 +269,9 @@ extension BaseTestCase {
         settings.cells.staticTexts["Passwords & Accounts"].tap()
         settings.cells.staticTexts["AutoFill Passwords"].tap()
         waitforExistence(settings.switches["AutoFill Passwords"], timeout: 3)
-        settings.switches["AutoFill Passwords"].tap()
+        if let switchValue = settings.switches["AutoFill Passwords"].value as? Bool, switchValue == false {
```
jhugman: 2 nits: 

1. put the second clause on a new line
2. use `not switchValue`, or the equivalent. `switchValue` is already a `Bool`, so we don't need to compare with a value.

CredentialProvider/Presenter/CredentialWelcomePresenter.swift
```diff
@@ -66,7 +66,7 @@ class CredentialWelcomePresenter: BaseWelcomePresenter {
     func onViewAppeared() {
         self.authenticationBag = DisposeBag()
 
-        let delay = isRunningTest ? 0.0 : 1.0
+        let delay = isRunningTest ? DispatchTimeInterval.seconds(0) : DispatchTimeInterval.seconds(1)
```
jhugman: Nit: consider 

```swift
let delay: DispatchTimeInterval = isRuningTest ? .seconds(0) : .seconds(1)
```

Shared/Presenter/BaseItemListPresenter.swift
```diff
@@ -150,15 +150,15 @@ extension BaseItemListPresenter {
                                           itemDetailIdObservable: Observable<String>,
                                           sidebarObservable: Observable<Bool>) -> Driver<[ItemSectionModel]> {
         // only run on a delay for UI purposes; keep tests from blocking
-        let listThrottle = isRunningTest ? 0.0 : 1.0
-        let stateThrottle = isRunningTest ? 0.0 : 2.0
+        let listThrottle = isRunningTest ? DispatchTimeInterval.seconds(0) : DispatchTimeInterval.seconds(1)
+        let stateThrottle = isRunningTest ? DispatchTimeInterval.seconds(0) : DispatchTimeInterval.seconds(2)
```
jhugman: Nit: again, consider putting the type label on the variable, and use the `.seconds` notation.

lockbox-ios/Presenter/ItemListPresenter.swift
```diff
@@ -227,7 +227,7 @@ extension ItemListPresenter {
         Observable.combineLatest(self.dataStore.syncState, isManualRefreshObservable)
                 .map { SyncStateManual(syncState: $0.0, manualSync: $0.1.refreshing) }
                 .asDriver(onErrorJustReturn: SyncStateManual(syncState: .Synced, manualSync: false))
-                .throttle(2.0)
+                .throttle(DispatchTimeInterval.seconds(2))
```
jhugman: I think we can let type inference do its thing here. 

```swift
.throttle(.seconds(2))
```

Shared/View/BaseItemListView.swift
```diff
@@ -84,21 +84,29 @@ class BaseItemListView: UIViewController {
...
-        self.navigationController?.navigationBar.tintColor = UIColor.white
-
-        self.navigationItem.title = Constant.string.productName
-        self.navigationController?.navigationBar.tintColor = UIColor.white
-        self.navigationController?.navigationBar.accessibilityIdentifier = "firefoxLockwise.navigationBar"
-        self.navigationController?.navigationBar.titleTextAttributes = [
+        navigationItem.title = Constant.string.productName
+        navigationController?.navigationBar.tintColor = UIColor.white
+        navigationController?.navigationBar.accessibilityIdentifier = "firefoxLockwise.navigationBar"
+        navigationController?.navigationBar.titleTextAttributes = [
```
jhugman: Consider putting this in an `if let`.

```swift
if let navigationBar = navigationController?.navigationBar {

    if #available(iOS 13.0, *) {
    }
}

```

jhugman: Also: group like things with like:

```swift
navigationItem.title = Constant.string.productName
navigationItem.largeTitleDisplayMode = .never
```

lockbox-iosTests/BaseDataStoreSpec.swift
```diff
@@ -43,8 +43,9 @@ class BaseDataStoreSpec: QuickSpec {
             self.ensureLockedCalled = true
         }
 
-        func sync(unlockInfo: SyncUnlockInfo) throws {
+        func sync(unlockInfo: SyncUnlockInfo) throws -> String {
             self.syncArgument = unlockInfo
+            return ""
```
jhugman: Nit: prefer a string that documents what this is, rather than an empty string. 

e.g. 

```swift
return "fake-unlock-info"
```

Shared/View/BaseItemListView.swift
```diff
@@ -84,21 +84,29 @@ class BaseItemListView: UIViewController {
     }
 
     internal func styleNavigationBar() {
```
jhugman: Are there any other places this method could go. Should it go in a common `UIViewController+` extension, and called as part of `viewDidLoad`?

kaylagalway: Yes - I would like to refactor all of the navigation styling in a later PR.

PR #1114, opened 2019-11-08
Add Feature Flags

kaylagalway:
Fixes [1095](https://github.com/mozilla-lockwise/lockwise-ios/issues/1095) 

## Testing and Review Notes
- This PR adds Feature Flags to the app for testing non-released features 

One example to test this functionality:
- Open app in Xcode 11.2.1
- Below is example of somewhere easy to change the code to test. This will change the navigation bar background color depending on feature flag status:

```
Navigate to extension UINavigationController and replace iosThirteenNavBarAppearance() function with this code instead:

func iosThirteenNavBarAppearance() {
        if #available(iOS 13.0, *) {
            let navBarAppearance = UINavigationBarAppearance()
            navBarAppearance.configureWithOpaqueBackground()
            navBarAppearance.titleTextAttributes = [.foregroundColor: UIColor.white]
            navBarAppearance.largeTitleTextAttributes = [.foregroundColor: UIColor.white]
            if FeatureFlags.crudEdit {
                navBarAppearance.backgroundColor = Constant.color.navBackgroundColor
            } else {
                navBarAppearance.backgroundColor = UIColor.red
            }
            navigationBar.standardAppearance = navBarAppearance
            navigationBar.scrollEdgeAppearance = navBarAppearance
        }
    }
```
- Then run the app normally in simulator (default is Debug) and you should see the normal purple navigation bar
- Click lockbox target in top left of Xcode screen and click "edit scheme"
- On left menu click "Run"
- Under "Info" tab change "Build Configuration" from Debug to Release
- Run the app again in simulator and you should see a red navigation bar

### DIFF
--- a/Shared/FeatureFlags/FeatureFlags.swift
+++ b/Shared/FeatureFlags/FeatureFlags.swift
@@ -1,16 +1,14 @@
 //
 //  FeatureFlags.swift
 //  Lockbox
-//
-//  Created by Kayla Galway on 11/8/19.
-//  Copyright © 2019 Mozilla. All rights reserved.
-//
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
 
 /// Helper struct that checks if Build Configuration is Release or Debug
 struct FeatureFlagSupport {
-    
     static var isDebug: Bool {
         #if DEBUG
             return true
@@ -26,15 +24,18 @@ struct FeatureFlagSupport {
             return false
         #endif
     }
-    
 }
 
 /// Boolean flags that identify if test feature should be shown depending on whether the Build Configuration is Release or Debug
 struct FeatureFlags {
-    
     /// Feature flag for the ability to edit a credential
     static var crudEdit: Bool {
-        return FeatureFlagSupport.isDebug ? true : false
+        if FeatureFlagSupport.isDebug {
+            return true
+        } else if FeatureFlagSupport.isRelease {
+            return false
+        } else {
+            return false
+        }
     }
-    
 }
--- a/Shared/FeatureFlags/FeatureFlags.swift
+++ b/Shared/FeatureFlags/FeatureFlags.swift
@@ -1,0 +1,40 @@
+//
+//  FeatureFlags.swift
+//  Lockbox
+//
+//  Created by Kayla Galway on 11/8/19.
+//  Copyright © 2019 Mozilla. All rights reserved.
+//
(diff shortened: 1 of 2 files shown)


Shared/FeatureFlags/FeatureFlags.swift
```diff
@@ -0,0 +1,40 @@
+//
+//  FeatureFlags.swift
+//  Lockbox
+//
+//  Created by Kayla Galway on 11/8/19.
+//  Copyright © 2019 Mozilla. All rights reserved.
+//
```
jhugman: Use the Mozilla standard copyright header please.

Shared/FeatureFlags/FeatureFlags.swift
```diff
@@ -0,0 +1,40 @@
...
+    }
+    
+    static var isRelease: Bool {
+        #if RELEASE
+            return true
+        #else
+            return false
+        #endif
+    }
+
```
jhugman: Nit: whitespace.

Shared/FeatureFlags/FeatureFlags.swift
```diff
@@ -0,0 +1,40 @@
...
+}
+
+/// Boolean flags that identify if test feature should be shown depending on whether the Build Configuration is Release or Debug
+struct FeatureFlags {
+    
+    /// Feature flag for the ability to edit a credential
+    static var crudEdit: Bool {
+        return FeatureFlagSupport.isDebug ? true : false
+    }
+
```
jhugman: Nit: whitespace.

Shared/FeatureFlags/FeatureFlags.swift
```diff
@@ -0,0 +1,40 @@
...
+    }
+    
+}
+
+/// Boolean flags that identify if test feature should be shown depending on whether the Build Configuration is Release or Debug
+struct FeatureFlags {
+    
+    /// Feature flag for the ability to edit a credential
+    static var crudEdit: Bool {
+        return FeatureFlagSupport.isDebug ? true : false
```
jhugman: I'm stuck here: do we make this simpler

e.g. 

```swift
static let crudEdit: Bool = FeatureFlagSupport.isDebug
```

or more verbose

```swift
static var crudEdit: Bool {
   switch {
       case FeatureFlag.isDebug: return true
       case FeatureFlag.isRelease: return false
   }

   return false
}
```

Having written this out, I think I'd prefer the latter as it's easier to change, and gives a good basis to copy from.

kaylagalway: You can't switch over the values of `FeatureFlagSupport` because it isn't an enum, so I used if statements to make it more verbose like your example.

PR #1130, opened 2019-11-13
Universal Links

kaylagalway:
Fixes #[1100](https://github.com/mozilla-lockwise/lockwise-ios/issues/1100)

## Testing and Review Notes

- Run the application on an iOS device
- Open the Notes application 
- Create new note with the following universal link: https://bhqf.adj.st
- Click that link and Lockwise should open


kaylagalway:
> Looks great! Have you rebased off of master on this branch? It seems like there's a lot of commits that are showing up 🤔 maybe try `git rebase master` and then `push -f` to this branch?

We chatted about this, but just wanted to note it here. This was branched off of ios-13-ui which was squash and merged into master, so this still shows all those commits.

PR #1134, opened 2019-11-15
Update release notes for release 1.7.0

kaylagalway:
Updates release notes for release 1.7.0
[Release 1.6.3 Ticket](https://github.com/mozilla-lockwise/lockwise-ios/issues/1107)

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -1,5 +1,19 @@
 # Release Notes
 
+## 1.6.3 (Build 3873)
+
+_Date: 2019-11-15_
+
+We fixed bugs related to: iOS 13 layout and navigation bar issues, keyboard dismissal when viewing login list
+
+All changes since last build and version 1.6.2:
+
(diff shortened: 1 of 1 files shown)


docs/release-notes.md
```diff
@@ -1,5 +1,19 @@
 # Release Notes
 
+## 1.6.3 (Build 3873)
+
+_Date: 2019-11-15_
+
+We fixed bugs related to: iOS 13 layout and navigation bar issues, keyboard dismissal when viewing login list
+
+All changes since last build and version 1.6.2:
+
```
devinreams: Mentioned via Slack that there may be a few more things coming along since the 1.6.2 release, seen in this comparison: https://github.com/mozilla-lockwise/lockwise-ios/compare/1.6.2.3616...master

And I believe the decision at the time was "swipe to delete" was a big enough feature to warrant bumping up to 1.7.0 (which is the current version in the CredentialProvider and shared Info.plists).

PR #1158, opened 2019-12-10
Remove sending of fxa id in telemetry events

kaylagalway:
Fixes [Buzilla Issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1602844)


Shared/Action/TelemetryAction.swift
```diff
@@ -126,14 +126,12 @@ class TelemetryActionHandler: ActionHandler {
...
         return Binder(self) { target, action in
-            let extras = self.addUidTo(extras: action.extras)
-
             target.telemetry.recordEvent(
                     category: TelemetryEventCategory.action.rawValue,
                     method: action.eventMethod.rawValue,
                     object: action.eventObject.rawValue,
                     value: action.value,
-                    extras: extras
+                    extras: nil
```
jhugman: This removes all extras from the event; is this what you intend?

Also: can we remove `addUidTo()` method?

PR #1160, opened 2019-12-11
Update localization strings

kaylagalway:
Pulls in new Localization strings


kaylagalway:
This build is currently failing because of the following error: `error: Build input file cannot be found: '/Users/kaylagalway/dev/mozilla/lockwise-ios/lockbox-ios/Storyboard/es.lproj/NoNetwork.strings' (in target 'Firefox Lockbox' from project 'Lockbox')`

Working on resolving this.

PR #1195, opened 2020-02-13
Fix Delete Button in production

### LINKED ISSUE #1188: Login is not deleted after tapping delete button
Steps to reproduce:
1. Install Lockwise on an iOS device.
2. Login into Lockwise.
3. Tap on a saved login.
4. Tap "Delete Login' button.
5. Observe.

Expected behavior
The saved login should be deleted

Actual behavior
The saved login is not deleted after tapping the 'Delete Login' button

* Device: iPhone Xs Max; iPad Air 3
* Build version: Lockwise master branch v1.7.2 (4059)

kaylagalway:
Fixes #1188 

- Delete button was working in debug mode, but not in production
- Delete subscription code was hidden behind edit feature flag, causing delete button not to work. Moved code outside of feature flag restrictions

### DIFF
--- a/lockbox-ios/Presenter/ItemDetailPresenter.swift
+++ b/lockbox-ios/Presenter/ItemDetailPresenter.swift
@@ -143,26 +143,26 @@ class ItemDetailPresenter {
     
     private func setupDelete() {
         view?.deleteTapped
-        .subscribe(onNext: { (_) in
-            self.view?.displayAlertController(
-                buttons: [
-                    AlertActionButtonConfiguration(
-                        title: Constant.string.cancel,
-                        tapObserver: nil,
-                        style: .cancel
-                    ),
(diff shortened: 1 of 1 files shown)


lockbox-ios/Presenter/ItemDetailPresenter.swift
```diff
@@ -136,9 +136,34 @@ class ItemDetailPresenter {
...
             setupEdit()
         }
+        setupDelete()
         setupCopy(itemObservable: itemObservable)
         setupNavigation(itemObservable: itemObservable)
     }
+    
+    private func setupDelete() {
+        view?.deleteTapped
+        .subscribe(onNext: { (_) in
```
jhugman: Nit: Whitespace

PR #1202, opened 2020-04-03
Resolves database access issues if locally stored salt or key is incorrect

### LINKED ISSUE #1201: Lockwise block on purple screen during launch
## Steps to reproduce
1. Launch Lockwise

### Expected behavior
List of password appears after a short loading

### Actual behavior
After the splash screen, the app blocks (soft lock) on a purple screen. Killing and relaunching the app doesn't help.

See screenshot below.

### Device & build information
* Device: iPhone 5s, latest OS (12.4.5)
* Build version: 1.7.2
### Notes
I don't even reach the part where Lockbox ask for my Touch ID. The app freezes before that.

Attachments:

kaylagalway:
Fixes #1201 

Database salt and plain text header migration failed for certain users with the update of 1.7.2, which left their app in an unusable state as the database was not accessible. This PR resolves those issues and resets the local database without deleting any user sync data.

### DIFF
--- a/Shared/Store/BaseDataStore.swift
+++ b/Shared/Store/BaseDataStore.swift
@@ -510,16 +510,19 @@ extension BaseDataStore {
         if keychainWrapper.hasValue(forKey: saltKey, withAccessibility: .afterFirstUnlock) {
             keychainWrapper.removeObject(forKey: saltKey)
         }
-        guard let database = loginsStorage as? LoginsStorage else { return nil }
-        database.close()
         do {
+            if let database = loginsStorage as? LoginsStorage {
+                database.close()
+            }
             if FileManager.default.fileExists(atPath: databasePath) {
                 try FileManager.default.removeItem(atPath: databasePath)
                 loginsStorage = nil
-                return createNewDatabase()
+                let newSalt = try createNewDatabase()
+                return newSalt
             } else {
                 loginsStorage = nil
-                return createNewDatabase()
+                let newSalt = try createNewDatabase()
+                return newSalt
             }
         } catch {
             self.dispatcher.dispatch(action: SentryAction(title: "handleDatabaseAccessFailure failed", error: error, line: nil))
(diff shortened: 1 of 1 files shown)


kaylagalway:
Build currently failing because new certificates need to be uploaded to buddybuild.  Resolving on Monday.

Shared/Store/BaseDataStore.swift
```diff
@@ -388,13 +388,17 @@ extension BaseDataStore {
...
+            let loginsDatabasePath = loginsDatabasePath else { return }
 
         do {
             try loginsStorage.ensureUnlockedWithKeyAndSalt(key: loginsKey, salt: salt)
             self.storageStateSubject.onNext(.Unlocked)
         } catch let error as LoginsStoreError {
             pushError(error)
+            // If we can not access database with current salt and key, need to delete local database and migrate to replacement salt
+            // This only deletes the local database file, does not delete the user's sync data
+            _ = handleDatabaseAccessFailure(databasePath: loginsDatabasePath, encryptionKey: loginsKey)
```
eliserichards, line 401: 👍

Shared/Store/BaseDataStore.swift
```diff
@@ -487,19 +491,59 @@ extension BaseDataStore {
...
+    // Closes database
+    // Deletes database file
+    // Creates new database and syncs
+    private func handleDatabaseAccessFailure(databasePath: String, encryptionKey: String) -> String? {
+        let saltKey = KeychainKey.salt.rawValue
+        if keychainWrapper.hasValue(forKey: saltKey, withAccessibility: .afterFirstUnlock) {
+            keychainWrapper.removeObject(forKey: saltKey)
+        }
+        guard let database = loginsStorage as? LoginsStorage else { return nil }
+        database.close()
```
eliserichards, line 514: What happens if we can't close the database properly here and we `return nil`? Should we attempt to `createNewDatabase()` if we don't have `loginsStorage`?

Shared/Store/BaseDataStore.swift
```diff
@@ -487,19 +491,63 @@ extension BaseDataStore {
...
+            }
+        } catch {
+            self.dispatcher.dispatch(action: SentryAction(title: "handleDatabaseAccessFailure failed", error: error, line: nil))
+        }
+    }
+    
+    enum DatabaseError: Error {
+        case issueDeletingDatabase(description: String)
+        case issueCreatingDatabase(description: String)
+    }
```
eliserichards, line 533: 👏

Shared/Store/BaseDataStore.swift
```diff
@@ -487,19 +491,63 @@ extension BaseDataStore {
...
+    
+    private func createNewDatabase() throws {
+        guard let encryptionKey = loginsKey else { throw DatabaseError.issueCreatingDatabase(description: "logins database key is nil") }
```
eliserichards, line 536: 🔥

Shared/Store/BaseDataStore.swift
jhugman: If we fail in this method, what happens? e.g. if we throw an unexpected runtime exception, or if the app can't continue, perhaps we could fail more loudly (e.g. alert) than a crash?

kaylagalway: If this fails - maybe I could fire off a reset action from within the handleDatabaseAccessFailure function?  I actually changed this so that handleDatabaseAccessFailure doesn't return, as it didn't need to since the salt gets stored locally in the keychain.

PR #1244, opened 2021-01-26
Upgrade to Xcode 11.7

st3fan:
This patch upgrades to Xcode 11.7. Right now that only means upgrade to a new release of application-services 0.59.1 that was also compiled against Xcode 11.7.

I just realized that Glean is also a Rust library that may need a new release with an Xcode upgrade. What do you think @travis79


travis79:
Glean is offered as part of the app services megazord for iOS, but it's my understanding that only Firefox iOS consumes it from there.  Let me check a couple of things and see what it will take to update lockwise Glean and I'll get back to you tomorrow.

PR #1247, opened 2021-02-07
Fixes Bugzilla issue 1657178

st3fan:
This patch fixes Bugzilla issue 1657178.


SimonBasca:
Verified fix on 1.8.0 (4249).

